PR #55, opened 2026-05-10
feat(migrate): transport-aware MCP model with VS Code servers/inputs schema

### LINKED ISSUE #41: [feature-research] Modernize MCP migration for VS Code’s transport-aware server schema
## Focus Area
`mcp-ecosystem` — picked by random rotation (`25207955516 mod 10 = 6`).

## Gap Matrix

| Agent / Area | Upstream Surface | Currently Synced? | Proposed Action | Priority |
| ------------ | ---------------- | ----------------- | --------------- | -------- |
| VS Code MCP migration | `mcp.json` with top-level `servers`/`inputs`, stdio + HTTP/SSE transports, headers, sandbox, and dev-mode fields | Partial (vault round-trip only) | Add a first-class VS Code MCP parser/serializer instead of treating VS Code as legacy `mcpServers` JSON | High |
| MCP transport model | Streamable HTTP plus transport-level authorization and discovery metadata | No | Replace the stdio-only intermediate model with a transport-aware discriminated union | High |
| Claude/Cursor MCP translators | JSON migration to/from richer MCP configs | Partial (stdio-style fields only) | Preserve transport/auth metadata where possible and emit explicit incompatibility warnings otherwise | Medium |
| Codex MCP serializer | TOML emission from migrated MCP configs | Partial (stdio-style fields only) | Extend TOML serialization to support non-stdio metadata or preserve structured extras for future support | Medium |

## Finding
AgentSync already has the right file surface for VS Code MCP configs: `AgentPaths.vscode.mcpJson` resolves the profile-level `mcp.json`, and the VS Code adapter snapshots/applies that file as an opaque JSON artifact (`src/config/paths.ts:56-65`, `src/agents/vscode.ts:18-46`). The gap is in **migration**, not raw vault sync. `readSourceArtefacts()` treats VS Code `mcp.json` as a generic MCP source (`src/migrate/migrate.ts:70-85`), but the migration contract still assumes JSON-based MCP configs are shaped like `{ "mcpServers": { name: { command, args, env } } }` and nothing more (`src/migrate/translators/mcp.ts:4-31`).

That legacy contract is reinforced in tests: the MCP translator fixtures only cover `mcpServers` plus `command`/`args`/`env`, and the VS Code adapter tests do the same (`src/migrate/__tests__/translators/mcp.test.ts:4-27`, `src/agents/__tests__/vscode.test.ts:48-64`). On the write path, `applyMigrated()` merges JSON MCP content under `existingParsed.mcpServers` for every non-Codex target, including VS Code (`src/migrate/migrate.ts:173-201`). That no longer matches VS Code’s documented MCP schema, which now uses top-level `servers` and optional `inputs`, and supports transport-specific fields such as `type`, `url`, `headers`, `envFile`, `sandboxEnabled`, `sandbox`, and `dev`.

This mismatch matters because the broader MCP ecosystem is no longer “stdio tuple only.” The MCP spec now defines both stdio and Streamable HTTP transports, and the authorization spec layers OAuth-style discovery and bearer-token auth onto HTTP transports. VS Code already exposes that richer model in product docs, including remote HTTP servers, authentication headers, secure input variables, and sandbox policies. AgentSync’s current intermediate model cannot represent those transport/auth capabilities, so modern MCP configurations can be round-tripped within the same file format but cannot be migrated across agents without dropping semantics or generating the wrong shape.

## Rationale
This is a leverage point for AgentSync’s positioning as a cross-agent bridge. Users increasingly discover MCP servers in one client first — especially VS Code’s fast-moving MCP UX — and then expect to reuse that setup across Claude, Cursor, and Codex. If AgentSync keeps a stdio-only migration model, it will look like MCP “support” exists while silently excluding newer remote-server, auth, and sandbox workflows that are becoming normal upstream.

The cost of inaction is mostly correctness and trust: migrated configs will either fail to translate, lose auth-related metadata, or be written in a shape that the target client does not actually use. Fixing the model now is feasible inside the current architecture because the adapter registry, per-agent path map, and pure translator registration system are already in place; the missing piece is a richer MCP data model plus target-aware serializers.

## References
- **Internal**:
  - `src/config/paths.ts:56-65`
  - `src/agents/vscode.ts:18-46`
  - `src/migrate/migrate.ts:70-85`
  - `src/migrate/migrate.ts:173-201`
  - `src/migrate/translators/mcp.ts:4-31`
  - `src/migrate/__tests__/translators/mcp.test.ts:4-27`
  - `src/agents/__tests__/vscode.test.ts:48-64`
- **External**:
  - https://code.visualstudio.com/docs/copilot/reference/mcp-configuration
  - https://code.visualstudio.com/docs/copilot/chat/mcp-servers
  - https://modelcontextprotocol.io/specification/2025-03-26/basic/transports
  - https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization
  - https://modelcontextprotocol.io/specification/2025-11-25/changelog

## Suggested Implementation Sketch
- Define a transport-aware MCP intermediate model in `src/migrate/translators/mcp.ts` using a discriminated union for at least `stdio`, `http`, and `sse`, plus optional auth/header/input metadata.
- Add a dedicated VS Code parser/serializer for `mcp.json` that understands `servers`, `inputs`, sandbox fields, and dev-mode fields instead of reusing the legacy `mcpServers` JSON parser.
- Update the JSON write path in `src/migrate/migrate.ts` so MCP merges are target-specific rather than always writing `mcpServers` for every non-Codex target.
- Preserve unsupported-but-known fields during migration whenever the target format can carry them; when it cannot, surface explicit warnings instead of silently dropping them.
- Extend translator and adapter tests with official-shape VS Code fixtures and transport-aware cases (HTTP/SSE, headers, input placeholders, sandbox settings).
- Keep raw vault snapshot/apply behavior intact so this change extends migration correctness without breaking existing same-agent sync flows.

*Generated by Feature Opportunity Researcher run #25207955516*

> Generated by [Feature Opportunity Researcher](https://github.com/chrisleekr/agentsync/actions/runs/25207955516/agentic_workflow) · [◷](https://github.com/search?q=repo%3Achrisleekr%2Fagentsync+is%3Aissue+%22gh-aw-workflow-call-id%3A+chrisleekr%2Fagentsync%2Ffeature-research%22&type=issues)
> - [x] expires  on May 8, 2026, 8:29 AM UTC

chrisleekr:
## Summary

Closes #41.

Models MCP server configuration as a transport-aware intermediate (`stdio` | `http` | `sse`) so VS Code's official top-level `servers` + `inputs` schema round-trips correctly through every translator pair. Previously the migrate pipeline assumed a flat `mcpServers` shape, which silently dropped HTTP/SSE transport metadata, `headers`, `envFile`, `sandbox`, and top-level `inputs` when targeting VS Code.

## Changes

**`src/migrate/translators/mcp.ts`** — rewritten around a discriminated-union `McpServer` model. New parsers (`parseVsCodeMcp` with legacy `mcpServers` fallback, `parseMcpServersJson`, `parseCodexMcp`) emit a shared `McpModel`; new serializers (`serializeVsCodeMcp`, `serializeMcpServersJson`, `serializeCodexMcp`) project it back into each agent's native shape. Non-stdio servers dropped while targeting stdio-only agents (Claude/Cursor) now emit explicit translator warnings naming the server and the unsupported transport. Top-level `inputs` are preserved into VS Code and warned about for all other targets.

**`src/migrate/types.ts`** — `Translator` return type extended with optional `warnings: string[]` so lossy translations can surface guidance instead of silently dropping fields.

**`src/migrate/migrate.ts`** — `applyMigrated()` now merges using a target-aware key: VS Code merges into top-level `servers` and dedupes `inputs`, Claude/Cursor keep their `mcpServers` merge, Codex keeps its `[mcp.servers.*]` TOML merge. Translator warnings are propagated into `MigrateResult.warnings` with a `${from} → ${target} (${type}): …` prefix so the CLI surfaces them.

**Tests** — `src/migrate/__tests__/translators/mcp.test.ts` adds an official-shape VS Code fixture plus the legacy fallback fixture and covers: vscode↔codex transport round-trip, claude/cursor → vscode emitting top-level `servers` (not double-nested), vscode → claude/cursor warning about dropped http/sse servers and dropped `inputs`, vscode → codex preserving `headers`/`url`/`type` under `[mcp.servers.*]`. `src/migrate/__tests__/migrate.test.ts` adds orchestrator-level tests for cross-shape merge keys and warning propagation. `src/agents/__tests__/vscode.test.ts` adds a regression test confirming `headers.Authorization` values in VS Code official-shape `mcp.json` are still redacted by the existing sanitizer (which walks any JSON shape).

**`docs/migrate.md`** — new "MCP Transport Support" section documents the per-target transport matrix and how warnings surface; "Key Behaviours" notes the target-specific MCP merge key.

### Architecture

```mermaid
flowchart LR
    classDef source fill:#1a5276,color:#ffffff
    classDef model fill:#1e8449,color:#ffffff
    classDef target fill:#b9770e,color:#ffffff
    classDef warn fill:#922b21,color:#ffffff

    VS["VS Code mcp.json<br/>servers + inputs"]:::source
    CL["Claude mcp.json<br/>mcpServers"]:::source
    CR["Cursor mcp.json<br/>mcpServers"]:::source
    CX["Codex config.toml<br/>[mcp.servers.*]"]:::source

    MODEL["McpModel<br/>discriminated union:<br/>stdio | http | sse<br/>+ inputs"]:::model

    OUTVS["VS Code: top-level servers + inputs"]:::target
    OUTCL["Claude: mcpServers (stdio-only)"]:::target
    OUTCR["Cursor: mcpServers (stdio-only)"]:::target
    OUTCX["Codex: [mcp.servers.*] TOML"]:::target

    WARN["MigrateResult.warnings<br/>names dropped server + transport"]:::warn

    VS --> MODEL
    CL --> MODEL
    CR --> MODEL
    CX --> MODEL

    MODEL --> OUTVS
    MODEL -->|"drop non-stdio"| OUTCL
    MODEL -->|"drop non-stdio"| OUTCR
    MODEL -->|"drop inputs"| OUTCX

    OUTCL -.-> WARN
    OUTCR -.-> WARN
    OUTCX -.-> WARN
```

## Test Evidence

| Command | Result |
|---------|--------|
| `bun run typecheck` | pass |
| `bun run lint` | pass (warnings only, no errors) |
| `bun test src/migrate src/agents/__tests__/vscode.test.ts` | **73 pass / 0 fail** |

Plan tasks T1–T11 all landed in the single commit on this branch. Full-suite `bun test` shows the 20 pre-existing `node:fs/promises` import failures from the daemon installer tests bleeding into unrelated suites; they are present on `main` at `72de7b1` without these changes and are not introduced by this PR (`migrate.test.ts` already carries a defensive workaround comment for that bleed).

## Risks / Follow-ups

- Codex serialization now emits structured `type` / `url` / `headers` / `auth` blocks under `[mcp.servers.<name>]` for http/sse transports. Existing Codex clients that only consume stdio fields will ignore them safely, but a future Codex release that itself defines a non-stdio schema may need a translator update.
- VS Code parser accepts both the official `servers` shape and the legacy `mcpServers` shape so existing vault snapshots continue to work; we should drop the legacy fallback once we're confident no vaults still carry the old shape.
- The 20 pre-existing test failures from the daemon installer `node:fs/promises` mock bleed remain — out of scope here but worth a dedicated cleanup.

## Checklist

- [x] New code has tests where appropriate
- [x] Documentation updated if behavior changed

## Summary by CodeRabbit

* **New Features**
  * Added support for MCP HTTP and SSE transports with intelligent platform-aware migration handling.
  * Introduced warning system for unsupported transports when migrating between platforms.
  * Enhanced secret redaction for MCP authorization headers and sensitive content.

* **Documentation**
  * Updated migration guide with MCP transport support details and platform-specific merge behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/55)

### DIFF
--- a/src/migrate/__tests__/migrate.test.ts
+++ b/src/migrate/__tests__/migrate.test.ts
@@ -11,7 +11,7 @@ import { createRequire } from "node:module";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { AgentPaths } from "../../config/paths";
-import { performMigrate, readSourceArtefacts } from "../migrate";
+import { applyMigrated, performMigrate, readSourceArtefacts } from "../migrate";
 
 // Re-install the real node:fs/promises in bun's module cache. The daemon
 // installer test files (installer-linux/macos/windows) stub fs/promises at
@@ -480,6 +480,71 @@ describe("performMigrate", () => {
     // Existing inputs preserved
     expect(Array.isArray(parsed.inputs)).toBe(true);
     expect((parsed.inputs as Array<{ id: string }>)[0].id).toBe("existing_token");
+  });
+
+  test("applyMigrated vscode inputs collision: source value wins on shared id", async () => {
+    // Exercises the inputs-dedup branch in applyMigrated. No registered
+    // translator pair currently produces inputs into a vscode target that
+    // already has inputs, but the dedup contract is still load-bearing
+    // defensive code — pin "source wins on collision" so it matches the
+    // sibling servers merge and docs/migrate.md.
+    writeFixture(
+      testVscode.mcpJson,
+      JSON.stringify({
+        servers: { existing: { type: "stdio", command: "old" } },
+        inputs: [{ id: "tok", type: "promptString", description: "OLD" }],
+      }),
+    );
+    const incoming = `${JSON.stringify(
+      {
+        servers: { gh: { type: "stdio", command: "gh-mcp" } },
+        inputs: [{ id: "tok", type: "promptString", description: "NEW" }],
+      },
+      null,
+      2,
+    )}\n`;
+
+    const artifact = await applyMigrated("vscode", "mcp", "mcp.json", incoming, false);
+    expect(artifact).not.toBeNull();
+    const { readIfExists } = await import("../../agents/_utils");
+    const written = await readIfExists(testVscode.mcpJson);
+    const parsed = JSON.parse(written as string) as {
+      servers: Record<string, unknown>;
+      inputs?: Array<{ id: string; description?: string }>;
+    };
+    // Source wins for the collision
+    expect(parsed.inputs?.find((i) => i.id === "tok")?.description).toBe("NEW");
+    // No duplicate entries — dedup still works
+    expect((parsed.inputs ?? []).filter((i) => i.id === "tok")).toHaveLength(1);
+    // Both servers present after merge
+    expect(parsed.servers.existing).toBeDefined();
+    expect(parsed.servers.gh).toBeDefined();
+  });
+
+  test("vscode → claude with only http servers does not create empty mcpServers file", async () => {
+    writeFixture(
+      testVscode.mcpJson,
+      JSON.stringify({
+        servers: {
+          remote: { type: "http", url: "https://example.com/mcp" },
+        },
+      }),
+    );
+
+    const result = await performMigrate({
+      from: "vscode",
+      to: "claude",
+      type: "mcp",
+      dryRun: false,
+    });
+
+    expect(result.errors).toHaveLength(0);
+    // Translator warned, but no file was written.
+    expect(result.warnings.length).toBeGreaterThan(0);
+    expect(result.migrated).toHaveLength(0);
+    const { readIfExists } = await import("../../agents/_utils");
+    const written = await readIfExists(testClaude.mcpJson);
+    expect(written).toBeNull();
   });
 
   test("partial write failure continues remaining artefacts", async () => {
--- a/src/migrate/__tests__/migrate.test.ts
+++ b/src/migrate/__tests__/migrate.test.ts
@@ -359,6 +359,129 @@ describe("performMigrate", () => {
     expect(result.migrated[0].targetPath).toContain("review.md");
   });
 
+  test("vscode → claude writes under mcpServers (not servers)", async () => {
+    writeFixture(
+      testVscode.mcpJson,
+      JSON.stringify({
+        servers: {
+          gh: { type: "stdio", command: "gh-mcp", args: [], env: {} },
+        },
+      }),
+    );
+
+    const result = await performMigrate({
+      from: "vscode",
+      to: "claude",
+      type: "mcp",
+      dryRun: false,
+    });
+
+    expect(result.errors).toHaveLength(0);
+    expect(result.migrated).toHaveLength(1);
+    const { readIfExists } = await import("../../agents/_utils");
+    const written = await readIfExists(testClaude.mcpJson);
+    expect(written).not.toBeNull();
+    const parsed = JSON.parse(written as string) as Record<string, unknown>;
+    expect(parsed.mcpServers).toBeDefined();
+    expect(parsed.servers).toBeUndefined();
+    expect((parsed.mcpServers as Record<string, unknown>).gh).toBeDefined();
+  });
+
+  test("claude → vscode writes under top-level servers (no nested mcpServers.servers)", async () => {
+    writeFixture(
+      testClaude.mcpJson,
+      JSON.stringify({
+        mcpServers: { gh: { command: "gh-mcp", args: [], env: {} } },
+      }),
+    );
+
+    const result = await performMigrate({
+      from: "claude",
+      to: "vscode",
+      type: "mcp",
+      dryRun: false,
+    });
+
+    expect(result.errors).toHaveLength(0);
+    expect(result.migrated).toHaveLength(1);
+    const { readIfExists } = await import("../../agents/_utils");
+    const written = await readIfExists(testVscode.mcpJson);
+    expect(written).not.toBeNull();
+    const parsed = JSON.parse(written as string) as Record<string, unknown>;
+    expect(parsed.servers).toBeDefined();
+    // Critically, the legacy mcpServers key must NOT appear on the VS Code target.
+    expect(parsed.mcpServers).toBeUndefined();
+    const servers = parsed.servers as Record<string, Record<string, unknown>>;
+    expect(servers.gh.type).toBe("stdio");
+    expect(servers.gh.command).toBe("gh-mcp");
+    // And there must be no double-nesting like servers.gh.servers or
+    // mcpServers.servers anywhere.
+    expect((servers.gh as { servers?: unknown }).servers).toBeUndefined();
+  });
+
+  test("vscode → claude HTTP transport produces translator warnings on result", async () => {
+    writeFixture(
+      testVscode.mcpJson,
+      JSON.stringify({
+        servers: {
+          remote: {
+            type: "http",
+            url: "https://example.com/mcp",
+            headers: { "X-Trace": "on" },
+          },
+          gh: { type: "stdio", command: "gh-mcp" },
+        },
+      }),
+    );
+
+    const result = await performMigrate({
+      from: "vscode",
+      to: "claude",
+      type: "mcp",
+      dryRun: false,
+    });
+
+    expect(result.errors).toHaveLength(0);
+    // Translator warnings about dropped HTTP transport should surface
+    expect(result.warnings.length).toBeGreaterThan(0);
+    const joined = result.warnings.join("\n");
+    expect(joined).toContain("remote");
+    expect(joined).toContain("http");
+  });
+
+  test("vscode existing inputs preserved during vscode-target merge", async () => {
+    // Existing VS Code file with both servers and inputs
+    writeFixture(
+      testVscode.mcpJson,
+      JSON.stringify({
+        servers: { existing: { type: "stdio", command: "old" } },
+        inputs: [{ id: "existing_token", type: "promptString", password: true }],
+      }),
+    );
+    writeFixture(testClaude.mcpJson, JSON.stringify({ mcpServers: { gh: { command: "gh-mcp" } } }));
+
+    const result = await performMigrate({
+      from: "claude",
+      to: "vscode",
+      type: "mcp",
+      dryRun: false,
+    });
+
+    expect(result.errors).toHaveLength(0);
+    const { readIfExists } = await import("../../agents/_utils");
+    const written = await readIfExists(testVscode.mcpJson);
+    const parsed = JSON.parse(written as string) as {
+      servers: Record<string, unknown>;
+      inputs?: unknown[];
+    };
+    // Incoming server merged, existing server preserved
+    expect(parsed.servers.gh).toBeDefined();
+    expect(parsed.servers.existing).toBeDefined();
+    // Existing inputs preserved
+    expect(Array.isArray(parsed.inputs)).toBe(true);
+    expect((parsed.inputs as Array<{ id: string }>)[0].id).toBe("existing_token");
+  });
+
   test("partial write failure continues remaining artefacts", async () => {
     mkdirSync(testClaude.commandsDir, { recursive: true });
     writeFileSync(join(testClaude.commandsDir, "review.md"), "# Review");
--- a/src/migrate/__tests__/translators/mcp.test.ts
+++ b/src/migrate/__tests__/translators/mcp.test.ts
@@ -26,6 +26,46 @@ args = []
 [mcp.servers.slack.env]
 `;
 
+const FIXTURE_VSCODE_OFFICIAL = JSON.stringify(
+  {
+    inputs: [{ id: "gh_token", type: "promptString", description: "GitHub token", password: true }],
+    servers: {
+      github: {
+        type: "stdio",
+        command: "gh-mcp",
+        args: ["serve"],
+        // biome-ignore lint/suspicious/noTemplateCurlyInString: VS Code input variable syntax
+        env: { GITHUB_TOKEN: "${input:gh_token}" },
+        envFile: ".env.local",
+        sandboxEnabled: true,
+        sandbox: { networking: false },
+        dev: { watch: ["**/*.ts"] },
+      },
+      remote: {
+        type: "http",
+        url: "https://example.com/mcp",
+        headers: { Authorization: "Bearer test", "X-Trace": "on" },
+      },
+      sse: {
+        type: "sse",
+        url: "https://example.com/sse",
+      },
+    },
+  },
+  null,
+  2,
+);
+
+const FIXTURE_VSCODE_LEGACY = JSON.stringify(
+  {
+    mcpServers: {
+      github: { command: "gh-mcp", args: ["serve"], env: { GITHUB_TOKEN: "test" } },
+    },
+  },
+  null,
+  2,
+);
+
 describe("MCP translators", () => {
   test("empty mcpServers returns null", () => {
     expect(translateMcp.claudeToCursor('{"mcpServers":{}}')).toBeNull();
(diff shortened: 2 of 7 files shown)


chrisleekr:

🔍 **Code review complete**, 7 files, +803/-96.

## Summary

PR #55 introduces a transport-aware MCP intermediate model (`stdio | http | sse`) so VS Code's official top-level `servers` + `inputs` schema round-trips through every translator pair. The model, parsers, and serializers are well-factored, translator warnings now propagate through `MigrateResult.warnings`, and `applyMigrated` correctly uses target-specific merge keys (`servers` for VS Code, `mcpServers` for Claude/Cursor, `[mcp.servers.*]` for Codex). Tests pass (73/73 in scope), typecheck clean. I'm flagging one real correctness bug in the codex serializer (extras nests deeper on each round-trip) and two minor consistency issues — none are blockers but the codex extras issue should be fixed before this lands so users with custom VS Code fields don't see configs degrade on repeated migrate cycles.

## What was checked

- Read in full: `src/migrate/translators/mcp.ts`, `src/migrate/migrate.ts`, `src/migrate/types.ts`, `src/migrate/__tests__/translators/mcp.test.ts`, `src/migrate/__tests__/migrate.test.ts`, `src/agents/__tests__/vscode.test.ts`, `docs/migrate.md`.
- Cross-referenced: `src/migrate/registry.ts` (translator wiring), `src/core/sanitizer.ts` (`redactSecretLiterals` recursion semantics), `src/agents/registry.ts` (agent enumeration).
- Ran `bun run typecheck` (clean) and `bun test src/migrate src/agents/__tests__/vscode.test.ts` (73 pass / 0 fail).
- Verified bug claims by running short scripts against the actual translator entry points: extras nesting over three codex round-trips, inputs collision behaviour via `applyMigrated`, and empty-output-with-warnings behaviour for all-http and inputs-only sources.

## Findings

### `[major]` src/migrate/translators/mcp.ts:393 — Codex serializer nests `extras` deeper on every round-trip

`serializeCodexMcp` writes the model's unknown-key bag as a nested subtable (`entry.extras = s.extras`). On re-parse `pickExtras` captures the key `"extras"` itself (not in `STDIO_KNOWN_KEYS`/`HTTP_KNOWN_KEYS`) back into the model's `extras`, so each `vscode → codex → vscode` cycle wraps the field one level deeper. Confirmed by running three cycles with a custom `customMeta` field — by cycle three the value is at `servers.github.extras.extras.extras.customMeta`. The VS Code serializer at `mcp.ts:311-313` already does the right thing by spreading extras inline; codex should match. **Fix:** spread `s.extras` inline in `serializeCodexMcp` (same pattern as VS Code), or add `"extras"` to both KNOWN_KEYS sets and unpack it back into the model in `parseSingleServer`.

### `[minor]` src/migrate/migrate.ts:217 — Inputs merge collision semantic disagrees with servers merge and docs

The dedup loop iterates `[existingInputs, incomingInputs]` so the first-seen id wins → existing inputs survive collisions. The servers merge two lines above uses `{ ...existingServers, ...incomingServers }` (incoming wins), and `docs/migrate.md:79` documents "source value wins on collision." I confirmed by calling `applyMigrated` with a colliding `id`: existing description was kept. Today the collision case is unreachable (only vscode emits `inputs` and vscode→vscode isn't a registered pair), but the asymmetry is misleading defensive code. **Fix:** swap the iteration order to `[incomingInputs, existingInputs]`.

### `[minor]` src/migrate/translators/mcp.ts:428 — Comment vs. code mismatch leads to empty target files

The inline comment at `mcp.ts:423-425` states the intent is to surface warnings *without* writing an empty `mcpServers` file. The code only returns `null` when both servers and warnings are empty; if there are warnings (e.g. all-http source dropped while targeting stdio-only Claude), it still returns `{ content: '{"mcpServers":{}}', warnings }`, which the orchestrator writes. I confirmed: an all-http VS Code source migrating to a fresh Claude target ends up creating a brand-new `~/.claude/.claude.json` containing `{"mcpServers":{}}`. Same for an inputs-only source → codex. **Fix:** extend the `Translator` return shape with a warnings-only/skip variant and propagate warnings from the orchestrator without calling `applyMigrated`, or align the comment with the actual behaviour.

## Reasoning

- **Why I did not flag the secret-detection initial fieldName as a bug.** `migrate.ts:331` calls `redactSecretLiterals(parsed, "mcpServers")` regardless of target shape (VS Code uses `servers`, Codex uses `[mcp.servers.*]`). The `fieldName` argument is only used to label the redaction placeholder and warning string, and the recursion immediately overrides it with each child key (so `Authorization: "ghp_..."` becomes `$AGENTSYNC_REDACTED_AUTHORIZATION`). Detection still fires correctly across all shapes — the cosmetic `mcpServers` label is harmless.
- **Why per-server silent drops are not flagged.** `parseSingleServer` returns `null` on a failed zod parse, and the callers drop nulls without surfacing a warning. This is a UX-only concern; the affected shape (a server entry missing both `command` and `url`) isn't realistically reachable from any of the four source schemas in production. Not worth blocking on.
- **Why the existing-toml param in `serializeCodexMcp` is not flagged.** The parameter is unused by current callers (the merge happens in `applyMigrated`), but the function still works correctly and there's no behavioural impact. Pure dead-code nit.
- **Sanitizer regression check.** Verified the new VS Code top-level `servers` shape with `headers.Authorization` still triggers `redactSecretLiterals` — the new test at `src/agents/__tests__/vscode.test.ts:67-95` exercises this with a github-style PAT, and I confirmed by reading `redactSecretLiterals` that the recursion is shape-agnostic.
- **Merge correctness for VS Code/codex/claude/cursor.** Each target now uses the correct top-level key (`servers` / `[mcp.servers.*]` / `mcpServers`), and the regression test at `migrate.test.ts:390-420` specifically asserts no double-nested `mcpServers.servers`. Good defence against the original bug from #41.

_cost: $3.2540 · turns: 34 · duration: 393s_

src/migrate/translators/mcp.ts
```diff
@@ -14,135 +25,460 @@ import * as TOML from "@iarna/toml";
...
+      entry.type = s.transport;
+      entry.url = s.url;
+      if (s.headers) entry.headers = s.headers as TOML.AnyJson;
+      if (s.auth !== undefined && s.auth !== null) entry.auth = s.auth as TOML.AnyJson;
+    }
+    if (s.envFile !== undefined) entry.envFile = s.envFile;
+    if (s.sandboxEnabled !== undefined) entry.sandboxEnabled = s.sandboxEnabled;
+    if (s.sandbox !== undefined && s.sandbox !== null) entry.sandbox = s.sandbox as TOML.AnyJson;
+    if (s.dev !== undefined && s.dev !== null) entry.dev = s.dev as TOML.AnyJson;
+    if (s.extras && Object.keys(s.extras).length > 0) entry.extras = s.extras as TOML.AnyJson;
```
chrisleekr: _⚠️ Potential issue_ | _🟠 Major_

**Codex serializer nests `extras` deeper on every round-trip.**

`serializeCodexMcp` writes unknown fields as a nested `entry.extras = s.extras` subtable (`src/migrate/translators/mcp.ts:393`), but `parseSingleServer` re-captures any unrecognized key into `extras` via `pickExtras(obj, STDIO_KNOWN_KEYS)` — and "extras" itself is **not** in `STDIO_KNOWN_KEYS` or `HTTP_KNOWN_KEYS`. So each `vscode → codex → vscode` cycle wraps the field one level deeper (`extras → extras.extras → extras.extras.extras`). I confirmed this by running three cycles locally: a custom VS Code field `customMeta` ends up nested three levels deep. The VS Code serializer already does the right thing at `mcp.ts:311-313` by spreading extras inline.

**Suggested fix:** Either spread `s.extras` inline at the same level as `command`/`args`/`env` in `serializeCodexMcp` (mirroring the VS Code serializer), or include `"extras"` in both `STDIO_KNOWN_KEYS` and `HTTP_KNOWN_KEYS` and unpack it back into the model in `parseSingleServer`.

src/migrate/migrate.ts
```diff
@@ -185,8 +188,38 @@ export async function applyMigrated(
...
+                  const id =
+                    entry && typeof entry === "object"
+                      ? ((entry as { id?: unknown }).id as string | undefined)
+                      : undefined;
+                  const dedupeKey = typeof id === "string" ? id : JSON.stringify(entry);
+                  if (seen.has(dedupeKey)) continue;
+                  seen.add(dedupeKey);
+                  merged.push(entry);
+                }
+              }
```
chrisleekr, line 221: _💭 Suggestion_ | _🟡 Minor_

**Inputs-merge collision semantic disagrees with the servers merge and the docs.**

The dedup loop at `src/migrate/migrate.ts:206-217` iterates `[existingInputs, incomingInputs]` and skips any id already in `seen`, so on collision the **existing** input wins. Meanwhile the servers merge two lines up uses `{ ...existingServers, ...incomingServers }` (incoming wins), and `docs/migrate.md:79` states "Overwrite on collision: If the target already has a matching entry, the source value wins." I exercised this directly via `applyMigrated` with an incoming input that shared an `id` with the existing one — the existing description ("OLD") was kept. The case is currently unreachable because no registered translator pair produces inputs into a target that already has inputs (only vscode emits `inputs`, and vscode→vscode isn't registered), but the asymmetry is confusing defensive code and a footgun the next time the matrix grows.

**Suggested fix:** Reverse the iteration order (`[incomingInputs, existingInputs]`) so source wins on collision, matching servers and the doc.

src/migrate/translators/mcp.ts
```diff
@@ -14,135 +25,460 @@ import * as TOML from "@iarna/toml";
...
+    try {
+      const model = parse(raw);
+      if (model.servers.length === 0 && model.inputs.length === 0) return null;
+      const { content, warnings } = serializeMcpServersJson(model);
+      // Even if every server was dropped, surface the warnings so the
+      // operator hears about it — but skip writing an empty mcpServers file.
+      const parsed = JSON.parse(content) as { mcpServers: Record<string, unknown> };
+      if (Object.keys(parsed.mcpServers).length === 0) {
+        if (warnings.length === 0) return null;
+        return { content, targetName: "mcp.json", warnings };
```
chrisleekr: _💭 Suggestion_ | _🟡 Minor_

**Comment says "skip writing an empty mcpServers file" but the code still writes one.**

The inline comment at `src/migrate/translators/mcp.ts:423-425` states the intent is to surface warnings *without* writing the empty file, but both branches of the `if (Object.keys(parsed.mcpServers).length === 0)` block return `{ content, targetName, warnings }` — the only `return null` is when warnings are also empty. The orchestrator at `migrate.ts:355-365` happily writes whatever it receives, so an all-http VS Code source migrating to a fresh Claude target creates a brand-new `~/.claude/.claude.json` containing `{"mcpServers":{}}` (I verified this). Same for an inputs-only source → codex producing a near-empty `[mcp]\nservers = { }\n` file.

**Suggested fix:** Make the `Translator` return shape support a warnings-only result (e.g. `{ warnings, skip: true }`) and have the orchestrator propagate the warnings without calling `applyMigrated`. Alternatively, align the comment with the actual behaviour if creating an empty target file is intentional.

chrisleekr:

## Outstanding

Nothing blocks merge. PR state: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, 0 unresolved review threads, 10/10 CI checks green. `reviewDecision` is empty (no formal review-approval workflow gated this PR).

Local note: the 20 pre-existing `node:fs/promises` mock-bleed failures called out in the PR description still exist on `main`; my changes do not introduce or remove them. Tracked as a separate cleanup per the PR's own follow-ups list.

_cost: $4.9999 · turns: 56 · duration: 577s_

src/migrate/translators/mcp.ts
chrisleekr: _✅ Addressed_, commit `cd2c259`

**Codex serializer now spreads `extras` inline like VS Code.**

Confirmed locally: `entry.extras = s.extras` re-captures the literal key `"extras"` via `pickExtras` on every parse, nesting one level deeper each round-trip. Fixed in `src/migrate/translators/mcp.ts` by spreading the bag at the same level as the typed fields (mirroring `serializeVsCodeMcp`). Added a 3-cycle regression test (`vscode → codex → vscode does not re-nest extras over repeated cycles`) using a `customMeta` field — it now stays at the top level of the server entry across all three cycles.

src/migrate/migrate.ts
chrisleekr: _✅ Addressed_, commit `cd2c259`

**Inputs dedup now iterates `[incomingInputs, existingInputs]` so source wins on collision.**

Brings `applyMigrated` in line with the spread-based servers merge (`{...existing, ...incoming}`) and the documented "source value wins" rule in `docs/migrate.md:79`. Added a direct `applyMigrated` test (`applyMigrated vscode inputs collision: source value wins on shared id`) that hand-crafts the colliding-`id` content and asserts the source description (`"NEW"`) survives — exercises the dedup branch even though no registered translator pair currently produces inputs into a vscode target with existing inputs.

src/migrate/translators/mcp.ts
chrisleekr: _✅ Addressed_, commit `cd2c259`

**Empty-target write now skipped via a new `skipWrite` field on the `Translator` return type.**

Extended `Translator` in `src/migrate/types.ts` with `skipWrite?: boolean`. `mcpToMcpServers` sets `skipWrite: true` when no stdio server survived the projection (computed from the model — drops the redundant `JSON.parse(content)` while we're here), and `mcpToCodex` sets it when only `inputs` remain. The orchestrator at `migrate.ts` records the warnings on `MigrateResult.warnings` then `continue`s past `applyMigrated`, so no empty-stub file ever lands on disk. Added an orchestrator test (`vscode → claude with only http servers does not create empty mcpServers file`) that confirms `result.warnings.length > 0`, `result.migrated.length === 0`, and `~/.claude/.claude.json` is `null` afterwards.

src/migrate/translators/mcp.ts
```diff
@@ -14,135 +25,460 @@ import * as TOML from "@iarna/toml";
...
+      env: s.env,
+    });
+  }
+  if (model.inputs.length > 0) {
+    warnings.push(
+      `Dropped ${model.inputs.length} top-level "inputs" placeholder(s): not supported by the target's mcpServers schema.`,
+    );
+  }
+  return { content: `${JSON.stringify({ mcpServers }, null, 2)}\n`, warnings };
+}
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `cd2c259`

`serializeMcpServersJson` now collects `envFile`, `sandbox`, `sandboxEnabled`, `dev`, and any preserved `extras` keys on stdio servers into a per-server warning (`Server "<name>": dropped fields not supported by mcpServers schema: ...`). Added a regression test (`vscode → claude warns about dropped stdio-only metadata fields`) using the existing `FIXTURE_VSCODE_OFFICIAL` which exercises all four fields on the `github` stdio server. Docs/migrate.md transport table updated to list these lossy fields.

src/migrate/translators/mcp.ts
```diff
@@ -14,135 +25,460 @@ import * as TOML from "@iarna/toml";
...
+    if (s.sandboxEnabled !== undefined) entry.sandboxEnabled = s.sandboxEnabled;
+    if (s.sandbox !== undefined && s.sandbox !== null) entry.sandbox = s.sandbox as TOML.AnyJson;
+    if (s.dev !== undefined && s.dev !== null) entry.dev = s.dev as TOML.AnyJson;
+    if (s.extras && Object.keys(s.extras).length > 0) entry.extras = s.extras as TOML.AnyJson;
+    serverMap[s.name] = entry;
   }
   mcp.servers = serverMap;
   base.mcp = mcp;
   return TOML.stringify(base);
 }
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `cd2c259`

Removed the `existingToml` parameter and its try/parse branch from `serializeCodexMcp` — merge ownership lives in `applyMigrated` and the only caller (`mcpToCodex`) was already passing zero args. Clarified the JSDoc to make the ownership explicit.

src/migrate/translators/mcp.ts
```diff
@@ -14,135 +25,460 @@ import * as TOML from "@iarna/toml";
...
+      if (Object.keys(parsed.mcpServers).length === 0) {
+        if (warnings.length === 0) return null;
+        return { content, targetName: "mcp.json", warnings };
+      }
+      return { content, targetName: "mcp.json", warnings };
+    } catch {
+      return null;
+    }
+  };
+}
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `cd2c259`

Subsumed by the `skipWrite` refactor for the chrisleekr `mcp.ts:428` thread. `mcpToMcpServers` now derives `wouldWriteAnyServer` directly from the model (`model.servers.some(s => s.transport === "stdio")`) — no `JSON.parse(content)` round-trip.

src/migrate/__tests__/translators/mcp.test.ts
```diff
@@ -76,3 +116,120 @@ describe("MCP translators", () => {
...
+    expect(parsed.servers.github.type).toBe("stdio");
+    expect(parsed.servers.github.command).toBe("gh-mcp");
+    expect(parsed.servers.remote.type).toBe("http");
+    expect(parsed.servers.remote.url).toBe("https://example.com/mcp");
+    expect((parsed.servers.remote.headers as Record<string, string>).Authorization).toBe(
+      "Bearer test",
+    );
+    expect(parsed.servers.sse.type).toBe("sse");
+    expect(parsed.servers.sse.url).toBe("https://example.com/sse");
+  });
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `cd2c259`

Comment updated to "Round-trip vscode → codex → vscode to prove parse+serialize work without losing transport metadata across the model."

src/migrate/__tests__/migrate.test.ts
```diff
@@ -359,6 +359,129 @@ describe("performMigrate", () => {
...
+      servers: Record<string, unknown>;
+      inputs?: unknown[];
+    };
+    // Incoming server merged, existing server preserved
+    expect(parsed.servers.gh).toBeDefined();
+    expect(parsed.servers.existing).toBeDefined();
+    // Existing inputs preserved
+    expect(Array.isArray(parsed.inputs)).toBe(true);
+    expect((parsed.inputs as Array<{ id: string }>)[0].id).toBe("existing_token");
+  });
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `cd2c259`

Added an explicit `applyMigrated` test (`applyMigrated vscode inputs collision: source value wins on shared id`) that hand-crafts incoming + existing content with the same input `id` and asserts (a) the source description survives, (b) only one entry exists for that `id` post-dedup. Done at the `applyMigrated` boundary because no registered translator pair currently produces inputs into a vscode target with pre-existing inputs.

PR #56, opened 2026-05-10
feat(cli): friendly errors for missing vault and second-machine pull

### LINKED ISSUE #44: CLI error UX: cryptic age error on second-machine pull, raw stack trace before init
## Summary

Two related CLI ergonomics gaps surface internal/library error wording instead of actionable guidance. Filing together because the fix is the same shape (catch known error → re-throw as a `log.error` with a hint).

## A. Second-machine onboarding fails with raw age error

**Repro** (against a local bare-repo remote):

```bash
# Machine A
HOME=/tmp/A AGENTSYNC_HOME=/tmp/A AGENTSYNC_VAULT_DIR=/tmp/A/vault \
  AGENTSYNC_KEY_PATH=/tmp/A/key.txt AGENTSYNC_MACHINE=laptop \
  bun run src/cli.ts init --remote /tmp/remote.git
# (push some content)

# Machine B (same remote, different machine)
HOME=/tmp/B AGENTSYNC_HOME=/tmp/B AGENTSYNC_VAULT_DIR=/tmp/B/vault \
  AGENTSYNC_KEY_PATH=/tmp/B/key.txt AGENTSYNC_MACHINE=desktop \
  bun run src/cli.ts init --remote /tmp/remote.git   # OK
bun run src/cli.ts pull
# ■  no identity matched any of the file's recipients
# exit 1
```

**Cause**: machine B's `init` correctly registers its pubkey in `agentsync.toml` and pushes that change, but it cannot re-encrypt existing `.age` artifacts (it has no decryption key). The codebase contract is: machine A must then run `agentsync key add <name-of-B> <pubkey-of-B>`, which `key.ts` uses to decrypt + re-encrypt every `.age` file for the union of recipients. **This contract is undocumented in any error or post-init message.**

**Proposed fix**:

1. After `init` joins an existing remote, print:
   > Your machine is now a recipient candidate. Hand this pubkey to a machine that can already decrypt the vault and ask them to run `agentsync key add <name> <pubkey>`. Until then, `pull` will fail.
2. In `pull.ts`, catch the age error containing `"no identity matched any of the file's recipients"` and rethrow with the same hint plus the local pubkey so it's copy-pasteable.

## B. `status` (and other vault-needing commands) before `init` print a raw Node stack trace

**Repro**:

```bash
HOME=/tmp/X AGENTSYNC_HOME=/tmp/X bun run src/cli.ts status
```

**Output** (excerpt):

```
3 | import { parse, stringify } from "@iarna/toml";
4 | import { type AgentSyncConfig, AgentSyncConfigSchema } from "./schema";
...
ENOENT: no such file or directory, open '/tmp/X/vault/agentsync.toml'
    path: "/tmp/X/vault/agentsync.toml",
 syscall: "open", errno: -2, code: "ENOENT"
   at async loadConfig (.../src/config/loader.ts:8:21)
   at async run (.../src/commands/status.ts:83:26)
exit=1
```

`doctor.ts` already wraps `loadConfig` in a try/catch that produces a clean check row. The same wrapper is missing from `status`, `push`, `pull`, `key`, `skill`.

**Proposed fix**: add a shared `loadVaultConfigOrExit()` helper in `commands/shared.ts` that calls `loadConfig` and on `ENOENT` prints:

> Vault not initialized at `<path>`. Run `agentsync init --remote <git-url>` first.

…and exits with code 1, with no stack.

## Out of scope

Not changing the `pull` cryptographic flow, only the surrounding error UX.

## Acceptance

- [ ] Fresh `init` on second machine prints recipient guidance.
- [ ] `pull` failing with the age-recipient error rethrows with hint.
- [ ] `status` / `push` / `pull` / `key` / `skill` before `init` print friendly message + exit 1, no stack.
- [ ] Tests cover both error paths.

chrisleekr:
## Summary

Fixes the two CLI UX papercuts called out in #44:

- **A. Cryptic age error on second-machine pull.** When a fresh machine ran `agentsync pull`, the age library threw the raw "no identity matched any of the file's recipients" error. `pull` now catches that signature and prints the recipient-handoff command the user must run on an existing machine, with the new machine's pubkey already filled in. `init` emits the same hint up-front so users learn what to do before they hit the failure.
- **B. Raw `ENOENT` stack before `init`.** Commands that read `agentsync.toml` (`status`, `push`, `pull`, `key`, `skill`) printed a Node stack trace if the vault directory did not exist yet. A new shared helper `loadVaultConfigOrExit` translates the missing-config case into a single-line friendly error pointing at `agentsync init --remote <url>` and exits 1.

Closes #44.

### `src/commands/shared.ts`
- Add `loadVaultConfigOrExit(vaultDir)` — wraps `loadConfig`, intercepts `ENOENT`, emits a friendly `log.error` line, exits with code 1. Non-ENOENT failures (schema / parse errors) re-throw so the existing error path stays intact.

### `src/commands/{status,push,pull,key,skill}.ts`
- Replace every `loadConfig(resolveConfigPath(runtime.vaultDir))` call site with `loadVaultConfigOrExit(runtime.vaultDir)`. `pull.ts` and `push.ts` also lift the config load out of any wrapping `try/catch` so the helper's `process.exit(1)` is what the user sees instead of an `errors` array entry.

### `src/commands/pull.ts`
- Detect the `no identity matched any of the file's recipients` substring in agent-apply failures, derive the local pubkey via age-encryption's `identityToRecipient`, and return a multi-line hint that includes the exact `agentsync key add <machine> <pubkey>` command.

### `src/commands/init.ts`
- When the machine is registering against an existing vault recipient set (not a fresh first-machine bootstrap), warn after `init` succeeds with the same recipient-handoff command so the next step is visible before the user tries `pull`.

### Tests
- `src/commands/__tests__/shared.test.ts` — direct coverage of the helper: ENOENT case prints the friendly message and exits 1; non-ENOENT (broken TOML) re-throws without calling `process.exit`.
- `src/commands/__tests__/pull.test.ts` — restructured to seed a real bare repo and machine fixture, then asserts the age recipient-mismatch path produces the hint with `agentsync key add <machine> <pubkey>`; non-age errors still propagate.
- `src/commands/__tests__/integration.test.ts` — covers `init` printing the recipient-onboarding warning when joining an existing vault, and not printing it on a clean first-machine bootstrap.

### Architecture

```mermaid
flowchart LR
    A[CLI command] --> B[resolveRuntimeContext]
    B --> C{loadVaultConfigOrExit}
    C -- config present --> D[continue command]
    C -- ENOENT --> E[log.error: Vault not initialized]
    E --> F[process.exit 1]
    D --> G[agent.apply per artifact]
    G -- age no identity --> H[buildRecipientHandoffHint]
    H --> I[errors: agentsync key add machine pubkey]
```

## Test Evidence

- `bun run check` — typecheck + biome clean; test suite reports 222 pass / 20 fail, matching the pre-change baseline (no new failures introduced).
- Targeted runs verified independently:
  - `bun test src/commands/__tests__/shared.test.ts` — 7/7 pass (3 existing + 2 new for `loadVaultConfigOrExit`).
  - `bun test src/commands/__tests__/pull.test.ts` — all pass including the new "performPull recipient-handoff error UX" describe (2 tests).
  - `bun test src/commands/__tests__/integration.test.ts` — all pass including the new init recipient-hint tests (2 tests).

## Risks / Follow-ups

- `loadVaultConfigOrExit` calls `process.exit(1)` directly. Test sites that exercise it stub `process.exit` to throw a sentinel; production callers see a single friendly line and exit code 1, which matches the rest of the CLI's error path.
- The age error signature is matched by substring (`no identity matched any of the file's recipients`). If the upstream library rewords its error message in a future release, the friendly hint will silently regress to the wrapped underlying message — the test in `pull.test.ts` will catch this on next bump.

## Checklist

- [x] New code has tests where appropriate
- [x] Documentation updated if behavior changed (user-facing strings only; no doc pages affected)

## Summary by CodeRabbit

* **New Features**
  * Added user guidance when joining an existing vault, directing users to run `agentsync key add` for proper access setup.

* **Bug Fixes**
  * Improved error handling when vault artifacts cannot be decrypted due to recipient mismatch, providing actionable guidance instead of generic errors.
  * Enhanced error messages for vault initialization and configuration loading issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/56)

### DIFF
--- a/src/commands/__tests__/integration.test.ts
+++ b/src/commands/__tests__/integration.test.ts
@@ -261,6 +261,51 @@ describe("integration", () => {
     );
   });
 
+  test("init prints the recipient-onboarding hint when joining a vault as a new machine", async () => {
+    const root = join(tmpDir, "recipient-handoff-hint");
+    mkdirSync(root, { recursive: true });
+    const bareRepoPath = await createBareRepo(root);
+    const machineA = await createMachineFixture(root, "machine-alpha");
+    const machineB = await createMachineFixture(root, "machine-beta");
+
+    seedVaultRepo({ machine: machineA, bareRepoPath });
+
+    await withMachineEnv(machineB, async () => {
+      await initMod.initCommand.run?.({
+        args: { remote: bareRepoPath, branch: "main" },
+        rawArgs: [],
+        cmd: {} as never,
+      } as never);
+    });
+
+    // The init flow should warn the new joiner that they cannot pull until an
+    // existing recipient runs `key add` for them. The message must include
+    // both the local machine name and its pubkey so it is copy-pasteable.
+    const hint = fakeLogs.warn.find((message) =>
+      message.includes("agentsync key add machine-beta"),
+    );
+    expect(hint).toBeDefined();
+    expect(hint).toContain(machineB.recipient);
+    expect(hint).toContain("`agentsync pull` on this machine will fail");
+  });
+
+  test("init does NOT print the recipient-onboarding hint on a fresh first-machine bootstrap", async () => {
+    const root = join(tmpDir, "first-machine-no-hint");
+    mkdirSync(root, { recursive: true });
+    const bareRepoPath = await createBareRepo(root);
+    const firstMachine = await createMachineFixture(root, "first-machine");
+
+    await withMachineEnv(firstMachine, async () => {
+      await initMod.initCommand.run?.({
+        args: { remote: bareRepoPath, branch: "main" },
+        rawArgs: [],
+        cmd: {} as never,
+      } as never);
+    });
+
+    expect(fakeLogs.warn.some((message) => message.includes("agentsync key add"))).toBe(false);
+  });
+
   test("init reports a controlled bootstrap failure when local history already diverged", async () => {
     const root = join(tmpDir, "init-divergence");
     mkdirSync(root, { recursive: true });
--- a/src/commands/__tests__/integration.test.ts
+++ b/src/commands/__tests__/integration.test.ts
@@ -369,6 +369,14 @@ describe("integration", () => {
       } as never);
     });
 
+    // Pin to the canonical lead phrase of the init hint (see src/commands/init.ts)
+    // so reworded variants like "key set" / "recipient add" still fall through to
+    // the negative assertion and don't silently bypass it.
+    expect(
+      fakeLogs.warn.some((message) =>
+        message.includes("This machine is registered but cannot decrypt"),
+      ),
+    ).toBe(false);
     expect(fakeLogs.warn.some((message) => message.includes("agentsync key add"))).toBe(false);
   });
 
--- a/src/commands/__tests__/integration.test.ts
+++ b/src/commands/__tests__/integration.test.ts
@@ -289,6 +289,72 @@ describe("integration", () => {
     expect(hint).toContain("`agentsync pull` on this machine will fail");
   });
 
+  test("recipient handoff: init + key add (idempotent on same pubkey) re-encrypts vault for new machine", async () => {
+    const root = join(tmpDir, "recipient-handoff-end-to-end");
+    mkdirSync(root, { recursive: true });
+    const bareRepoPath = await createBareRepo(root);
+    const machineA = await createMachineFixture(root, "machine-alpha");
+    const machineB = await createMachineFixture(root, "machine-beta");
+
+    seedVaultRepo({ machine: machineA, bareRepoPath });
+
+    // machineA pushes a real .age artifact encrypted ONLY for itself, so the
+    // resulting file is the exact thing machineB cannot read post-init.
+    const { encryptString, decryptString } = await import("../../core/encryptor");
+    const {
+      mkdir: mkdirAsync,
+      writeFile: writeFileAsync,
+      readFile: readFileAsync,
+    } = await import("node:fs/promises");
+
+    const machineAClaudeDir = join(machineA.vaultDir, "claude");
+    await mkdirAsync(machineAClaudeDir, { recursive: true });
+    const plaintext = "# handoff-secret\nbody\n";
+    const onlyAlphaCiphertext = await encryptString(plaintext, [machineA.recipient]);
+    await writeFileAsync(join(machineAClaudeDir, "HANDOFF.age"), onlyAlphaCiphertext, "utf8");
+    runGit(["add", "claude/HANDOFF.age"], machineA.vaultDir);
+    runGit(["commit", "-m", "seed: artifact for machine-alpha only"], machineA.vaultDir);
+    runGit(["push", "origin", "main"], machineA.vaultDir);
+
+    // machineB joins. init writes its pubkey into recipients and pushes.
+    await withMachineEnv(machineB, async () => {
+      await initMod.initCommand.run?.({
+        args: { remote: bareRepoPath, branch: "main" },
+        rawArgs: [],
+        cmd: {} as never,
+      } as never);
+    });
+
+    // machineA runs the exact command the onboarding hint suggests. After
+    // init, recipients["machine-beta"] already equals machineB.recipient, so
+    // a non-idempotent `key add` would fail with "already exists" here.
+    // The vault re-encryption is the whole point of this step.
+    runGit(["pull", "--ff-only", "origin", "main"], machineA.vaultDir);
+    fakeLogs.error.length = 0;
+    process.exitCode = 0;
+
+    await withMachineEnv(machineA, async () => {
+      await (keyMod.keyCommand.subCommands as unknown as Record<string, CommandDef>).add.run?.({
+        args: { name: "machine-beta", pubkey: machineB.recipient },
+        rawArgs: [],
+        cmd: {} as never,
+      } as never);
+    });
+
+    expect(process.exitCode).toBe(0);
+    expect(fakeLogs.error).toHaveLength(0);
+
+    // The artifact must now be decryptable by machineB's identity. Push the
+    // re-encrypted commit through, fetch on machineB, and decrypt directly.
+    runGit(["pull", "--ff-only", "origin", "main"], machineB.vaultDir);
+    const reEncrypted = await readFileAsync(
+      join(machineB.vaultDir, "claude", "HANDOFF.age"),
+      "utf8",
+    );
+    const decrypted = await decryptString(reEncrypted, machineB.identity);
+    expect(decrypted).toBe(plaintext);
+  });
+
   test("init does NOT print the recipient-onboarding hint on a fresh first-machine bootstrap", async () => {
     const root = join(tmpDir, "first-machine-no-hint");
     mkdirSync(root, { recursive: true });
--- a/src/commands/__tests__/pull.test.ts
+++ b/src/commands/__tests__/pull.test.ts
@@ -1,6 +1,7 @@
 import { afterAll, afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
-import { rm } from "node:fs/promises";
+import { rm, writeFile } from "node:fs/promises";
 import { createRequire } from "node:module";
+import { join } from "node:path";
 import {
   createBareRepo,
   createMachineFixture,
@@ -141,6 +142,22 @@ describe("performPull recipient-handoff error UX", () => {
     const result = await performPull({});
     expect(result.fatal).toBe(true);
     expect(result.errors[0]).toContain("disk full");
-    expect(result.errors[0] ?? "").not.toContain("agentsync key add");
+    expect(result.errors[0]).not.toContain("agentsync key add");
+  });
+
+  test("non-ENOENT config load failures surface as a friendly errors[] row, not a raw throw", async () => {
+    // Corrupt the seeded agentsync.toml so loadConfig's TOML parser throws.
+    // loadVaultConfigOrExit only swallows ENOENT into a friendly exit; any
+    // other failure must re-throw and be caught by performPull's outer try,
+    // becoming an errors[] row with fatal=true instead of crashing the CLI
+    // with a Node stack trace.
+    await writeFile(join(machine.vaultDir, "agentsync.toml"), "this is = not valid toml [", "utf8");
+
+    const result = await performPull({});
+
+    expect(result.fatal).toBe(true);
+    expect(result.errors).toHaveLength(1);
+    expect(result.errors[0]).toBeTruthy();
+    expect(result.errors[0]).not.toContain("agentsync key add");
   });
 });
--- a/src/commands/__tests__/pull.test.ts
+++ b/src/commands/__tests__/pull.test.ts
@@ -1,17 +1,60 @@
-import { afterEach, describe, expect, test } from "bun:test";
+import { afterAll, afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { rm } from "node:fs/promises";
+import { createRequire } from "node:module";
+import {
+  createBareRepo,
+  createMachineFixture,
+  createTmpDir,
+  seedVaultRepo,
+  type TestMachineFixture,
+} from "../../test-helpers/fixtures";
 import { __setPullAgentsForTesting, performPull } from "../pull";
 
+// Force the rest of this file to use the real fs/promises, undoing any
+// module-level fs mocks from earlier test files in the same Bun worker.
+{
+  const require = createRequire(import.meta.url);
+  // biome-ignore lint/style/useNodejsImportProtocol: deliberate alias to bypass mock cache
+  const realFsPromises = require("fs/promises") as typeof import("node:fs/promises");
+  mock.module("node:fs/promises", () => realFsPromises);
+}
+
+const RUNTIME_ENV_KEYS = ["AGENTSYNC_VAULT_DIR", "AGENTSYNC_KEY_PATH", "AGENTSYNC_MACHINE"];
+
 describe("performPull", () => {
-  afterEach(() => {
+  let tmpDir: string;
+  let machine: TestMachineFixture;
+  const savedEnv: Record<string, string | undefined> = {};
+
+  beforeEach(async () => {
+    tmpDir = await createTmpDir();
+    const bareRepoPath = await createBareRepo(tmpDir);
+    machine = await createMachineFixture(tmpDir, "pull-shape-machine");
+    seedVaultRepo({ machine, bareRepoPath, agents: { claude: true } });
+
+    for (const key of RUNTIME_ENV_KEYS) {
+      savedEnv[key] = process.env[key];
+    }
+    process.env.AGENTSYNC_VAULT_DIR = machine.vaultDir;
+    process.env.AGENTSYNC_KEY_PATH = machine.keyPath;
+    process.env.AGENTSYNC_MACHINE = machine.machineName;
+  });
+
+  afterEach(async () => {
     __setPullAgentsForTesting(null);
+    for (const key of RUNTIME_ENV_KEYS) {
+      const value = savedEnv[key];
+      if (value === undefined) delete process.env[key];
+      else process.env[key] = value;
+    }
+    await rm(tmpDir, { recursive: true, force: true });
   });
 
   test("accepts force option and returns valid result shape", async () => {
-    // performPull catches errors internally (never throws) and returns a result object.
-    // This test verifies:
-    //   1. TypeScript accepts { force: true } in the options (compile-time contract)
-    //   2. The function returns the expected result shape regardless of vault state
-    // The actual force→reconcileWithRemote behaviour is covered by git.test.ts.
+    // performPull never throws when called against a real vault; it returns
+    // a result object so the CLI can decide how to surface errors. We seed a
+    // vault explicitly because performPull's missing-vault branch now exits
+    // via loadVaultConfigOrExit rather than returning an error row.
     const result = await performPull({ force: true, dryRun: true });
 
     expect(result).toHaveProperty("applied");
@@ -20,5 +63,84 @@ describe("performPull", () => {
     expect(typeof result.applied).toBe("number");
     expect(Array.isArray(result.errors)).toBe(true);
     expect(typeof result.fatal).toBe("boolean");
+  });
+});
+
+describe("performPull recipient-handoff error UX", () => {
+  let tmpDir: string;
+  let machine: TestMachineFixture;
+  const savedEnv: Record<string, string | undefined> = {};
+
+  beforeEach(async () => {
+    tmpDir = await createTmpDir();
+    const bareRepoPath = await createBareRepo(tmpDir);
+    machine = await createMachineFixture(tmpDir, "second-machine");
+    seedVaultRepo({ machine, bareRepoPath, agents: { claude: true } });
+
+    for (const key of RUNTIME_ENV_KEYS) {
+      savedEnv[key] = process.env[key];
+    }
+    process.env.AGENTSYNC_VAULT_DIR = machine.vaultDir;
+    process.env.AGENTSYNC_KEY_PATH = machine.keyPath;
+    process.env.AGENTSYNC_MACHINE = machine.machineName;
+  });
+
+  afterEach(async () => {
+    __setPullAgentsForTesting(null);
+    for (const key of RUNTIME_ENV_KEYS) {
+      const value = savedEnv[key];
+      if (value === undefined) delete process.env[key];
+      else process.env[key] = value;
+    }
+    await rm(tmpDir, { recursive: true, force: true });
+  });
+
+  afterAll(() => {
+    mock.restore();
+  });
+
+  test("translates raw age recipient-mismatch into a copy-pasteable key add hint", async () => {
+    // Install a fake agent whose apply() raises the literal age library error
+    // a real second-machine pull would see. This proves the catch path in
+    // performPull recognises the marker without standing up a foreign-keyed
+    // .age artifact (which would require reaching into the age library
+    // internals to encrypt for someone else's recipient).
+    __setPullAgentsForTesting([
+      {
+        name: "claude",
+        snapshot: async () => ({ artifacts: [], warnings: [] }),
+        apply: async () => {
+          throw new Error("no identity matched any of the file's recipients");
+        },
+      },
+    ]);
+
+    const result = await performPull({});
+
+    expect(result.fatal).toBe(true);
+    expect(result.applied).toBe(0);
+    expect(result.errors).toHaveLength(1);
+    const message = result.errors[0] ?? "";
+    expect(message).toContain("agentsync key add");
+    expect(message).toContain(machine.machineName);
+    expect(message).toContain(machine.recipient);
+    expect(message).not.toMatch(/^no identity matched/);
+  });
+
+  test("non-age errors still propagate as-is so other failure modes stay visible", async () => {
+    __setPullAgentsForTesting([
+      {
+        name: "claude",
+        snapshot: async () => ({ artifacts: [], warnings: [] }),
+        apply: async () => {
+          throw new Error("disk full");
+        },
+      },
+    ]);
+
+    const result = await performPull({});
+    expect(result.fatal).toBe(true);
+    expect(result.errors[0]).toContain("disk full");
+    expect(result.errors[0] ?? "").not.toContain("agentsync key add");
   });
 });
--- a/src/commands/__tests__/pull.test.ts
+++ b/src/commands/__tests__/pull.test.ts
@@ -51,10 +51,10 @@ describe("performPull", () => {
   });
 
   test("accepts force option and returns valid result shape", async () => {
-    // performPull never throws when called against a real vault; it returns
-    // a result object so the CLI can decide how to surface errors. We seed a
-    // vault explicitly because performPull's missing-vault branch now exits
-    // via loadVaultConfigOrExit rather than returning an error row.
+    // performPull catches non-fatal failures internally and returns a result
+    // object so the CLI can decide how to surface errors; only the ENOENT
+    // missing-vault path exits via loadVaultConfigOrExit. We seed a vault so
+    // this run exercises the result-shape contract rather than the exit path.
     const result = await performPull({ force: true, dryRun: true });
 
     expect(result).toHaveProperty("applied");
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -107,6 +107,9 @@ export const initCommand = defineCommand({
         existing = null;
       }
 
+      const joinedExistingVault =
+        existing !== null && existing.recipients[runtime.machineName] !== recipient;
+
       const config = {
         version: existing?.version ?? "1",
         recipients: {
@@ -135,6 +138,20 @@ export const initCommand = defineCommand({
       }
 
       log.success(`Initialized vault at ${runtime.vaultDir}`);
+
+      if (joinedExistingVault) {
+        // The new machine wrote its pubkey into the recipient set but cannot
+        // re-encrypt existing .age artifacts itself (no decryption key). An
+        // existing recipient must run `key add` to grant read access.
+        log.warn(
+          [
+            "This machine is registered but cannot decrypt the existing vault yet.",
+            "An existing recipient must run on a machine that can decrypt the vault:",
+            `  agentsync key add ${runtime.machineName} ${recipient}`,
+            "Until that runs, `agentsync pull` on this machine will fail.",
+          ].join("\n"),
+        );
+      }
     } catch (err) {
       log.error(err instanceof Error ? err.message : String(err));
       process.exitCode = 1;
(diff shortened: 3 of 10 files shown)


chrisleekr:

🔍 **Code review complete**, 10 files, +339/-28.

## Summary

The PR's intent is solid — both UX papercuts (raw ENOENT before `init`; cryptic age error on second-machine `pull`) are real, and the helper extraction in `shared.ts` is clean and well-tested. **However, the new helpful hint guides users into a follow-up command (`agentsync key add <machine> <pubkey>`) that is guaranteed to fail.** I verified this empirically by driving `init` + `key add` against a real bare repo: `key add` errors with `"Recipient 'machine-beta' already exists"` because `init` already pushed the new machine's pubkey into the recipient set. The PR therefore replaces a cryptic error with a polite-but-broken instruction. Requesting changes on that, plus a test-coverage gap that allowed the broken hint to ship clean through CI.

## What was checked

Files read in full:
- `src/commands/shared.ts` (new `loadVaultConfigOrExit` helper)
- `src/commands/init.ts` (recipient-handoff warn)
- `src/commands/pull.ts` (age-error → handoff hint translation)
- `src/commands/push.ts`, `src/commands/key.ts`, `src/commands/skill.ts`, `src/commands/status.ts` (helper migration)
- `src/commands/__tests__/shared.test.ts`, `pull.test.ts`, `integration.test.ts` (new test coverage)
- `src/config/loader.ts` (to verify the ENOENT path)
- `src/core/encryptor.ts` (to verify `identityToRecipient` semantics in the hint builder)
- `src/test-helpers/fixtures.ts` (to understand the new pull-test seeding)

Cross-references performed:
- `grep loadConfig\(resolveConfigPath` → confirmed daemon (`src/daemon/index.ts:61, 71`) and doctor (`src/commands/doctor.ts:174`) still use raw `loadConfig`, but both wrap with their own context-specific error handling (daemon: `Startup failed: …` + exit 1; doctor: failed-check row in the diagnostic table), so the migration deliberately skipping them is correct.
- `grep "no identity matched"` → only one production constant, only one test reference, only one match site.
- Pre-PR `src/commands/pull.ts` to confirm the new per-agent try/catch doesn't regress the non-age error path (it doesn't — non-age errors are still surfaced via the outer catch, same as before).
- Verified the hint failure empirically with a one-off test exercising `initCommand` then `keyCommand.subCommands.add`, both run against a real bare repo and machine fixture. Test output captured: `KEY ADD ERROR: ["Recipient 'machine-beta' already exists. Use a different name or remove it first."]`, exit code 1.

Validation runs:
- `bun run lint` — 8 warnings, none in the PR's changed files (all CSS-related noise unrelated to this PR).
- `bun test src/commands/__tests__/shared.test.ts src/commands/__tests__/pull.test.ts` — 10 pass, 0 fail.

## Findings

[major] `src/commands/init.ts:150` — Suggested `agentsync key add` command is guaranteed to fail after `init`. By the time this warn fires, `init` has already added the new machine to `config.recipients` and pushed the commit (`src/commands/init.ts:113-138`). When the existing recipient runs `agentsync key add <machine> <pubkey>`, `key add` reconciles with the remote, reloads the config, and errors out at `src/commands/key.ts:90-96` with `"Recipient '<name>' already exists. Use a different name or remove it first."` Verified end-to-end against a real bare repo. The new hint replaces "cryptic age error" with "polite but incorrect instructions". **Recommended fix:** either (a) make `key add` idempotent when `recipients[name] === pubkey` (skip the duplicate-name error and proceed to re-encrypt the vault for the full recipient set), or (b) stop having `init` write the new machine into `recipients` itself and let `key add` be the single source of truth.

[major] `src/commands/pull.ts:105` — Same broken hint surfaces on the pull error path. Mirrors the `init.ts` issue. The `pull.test.ts` recipient-handoff coverage only asserts the message string is built; it doesn't drive the full `init → pull (fail) → key add (recover) → pull (success)` cycle, so the broken remediation flow passes CI silently. **Recommended fix:** resolve at the source (per the init finding) and add an end-to-end test that exercises the documented recovery sequence.

[minor] `src/commands/__tests__/integration.test.ts:290` — The new "init prints the recipient-onboarding hint" test asserts the warn text but never executes the suggested `key add` command. This is the test gap that let the broken hint ship clean. **Recommended fix:** extend (or sibling-add to) this test so it runs `key add machine-beta <pubkey>` on `machineA` after `machineB`'s init and asserts the vault re-encryption succeeds. That keeps the user-facing documented onboarding path end-to-end-tested.

## Reasoning

**Why the helper extraction is fine.** `loadVaultConfigOrExit` is a small, single-purpose function: catches only ENOENT, prints exactly the message the user needs (path + remediation command), exits 1. Other errors (TOML parse, Zod schema, EACCES) re-throw, so schema/parse failures still surface with full detail. The test in `shared.test.ts` covers both branches. I checked whether non-ENOENT IO errors (e.g. EACCES on a permission-restricted vault dir) would now produce a worse experience — they'd still produce a Node-shaped error, but that's the same as before this PR, so no regression and not worth flagging.

**Why I didn't flag the substring match for the age error.** The constant `AGE_NO_IDENTITY_MARKER = "no identity matched any of the file's recipients"` is brittle — if `age-encryption` rewords its error, this regresses silently to the raw underlying message. The PR description already calls this out as a known risk and the test in `pull.test.ts` would catch it on the next age-library bump. Acceptable risk given the alternative (subclassing age's error type or matching by error code) requires upstream changes.

**Why I didn't flag the daemon/doctor not being migrated.** The PR description says "Replace every `loadConfig(resolveConfigPath(...))` call site" but `src/daemon/index.ts:61, 71` and `src/commands/doctor.ts:174` are intentionally left alone — both already have their own context-specific error handling (daemon: friendly "Startup failed: …" + exit; doctor: diagnostic-table failed-check row). Migrating them to the shared helper would replace context-specific UX with the generic one. Correct call by the author.

**Why I didn't flag the per-agent try/catch in pull.ts.** Compared to the pre-PR single-outer-try structure, the new layout preserves the non-AGE error semantics (re-throws into the outer catch, ends up in `errors[]`), and adds correct early-return semantics for AGE (no point continuing — all agents share recipients). One subtle property: if multiple agents would fail with non-AGE errors, only the first surfaces — but that was also the behaviour before this PR, so no regression.

**Why the major findings are user-flow correctness rather than just "documentation".** The recipient-handoff hint is the entire payoff of fix "A" in this PR. If the suggested command doesn't work, the PR has succeeded at making the failure mode prettier but hasn't actually unblocked the second-machine pull workflow. That's why I flagged it as Major rather than Minor — it undermines the PR's stated goal.

_cost: $2.8947 · turns: 41 · duration: 356s_

src/commands/init.ts
```diff
@@ -135,6 +138,20 @@ export const initCommand = defineCommand({
...
+
+      if (joinedExistingVault) {
+        // The new machine wrote its pubkey into the recipient set but cannot
+        // re-encrypt existing .age artifacts itself (no decryption key). An
+        // existing recipient must run `key add` to grant read access.
+        log.warn(
+          [
+            "This machine is registered but cannot decrypt the existing vault yet.",
+            "An existing recipient must run on a machine that can decrypt the vault:",
+            `  agentsync key add ${runtime.machineName} ${recipient}`,
```
chrisleekr, line 150: _⚠️ Potential issue_ | _🟠 Major_

**Suggested `agentsync key add` command is guaranteed to fail with "Recipient already exists".**

By the time this hint is printed, `init` has already added the new machine to `config.recipients` and pushed the commit (`src/commands/init.ts:113-138`). When the existing recipient runs `agentsync key add <machineName> <pubkey>` on their machine, `key add` reconciles with the remote (`src/commands/key.ts:83-87`), reloads the config, and then errors out at `src/commands/key.ts:90-96` with `"Recipient '<name>' already exists. Use a different name or remove it first."` I verified this end-to-end by driving both commands against a real bare repo — the hint visibly fails. The same broken command is also surfaced in `src/commands/pull.ts:105`. So the new hint replaces "cryptic age error" with "polite but incorrect instructions", and the documented onboarding path now dead-ends without follow-up docs telling the user what to do next. The integration test at `src/commands/__tests__/integration.test.ts:264` only asserts the warn is emitted; it never executes the suggested command, so this regression isn't caught.

**Suggested fix:** Either (a) make `key add` idempotent when `recipients[name] === pubkey` (skip the duplicate-name error and proceed to re-encrypt vault files for the full recipient set), or (b) have `init` stop writing the new machine into `recipients` itself and let `key add` be the single source of truth. Then extend the integration test to actually run `key add` after the second-machine `init` and assert it succeeds and re-encrypts the vault.

src/commands/pull.ts
```diff
@@ -73,6 +85,28 @@ export async function performPull(
...
+  try {
+    const recipient = await identityToRecipient(identity);
+    pubkeyLine = recipient;
+  } catch {
+    // identity unreadable — fall back to generic guidance below.
+  }
+  return [
+    "Cannot decrypt vault: this machine's age key is not in the recipient set.",
+    "An existing machine that can decrypt the vault must add this machine as a recipient:",
+    `  agentsync key add ${machineName} ${pubkeyLine}`,
```
chrisleekr, line 110: _⚠️ Potential issue_ | _🟠 Major_

**Same broken `key add` hint surfaces on the pull error path.**

The hint here mirrors the one in `src/commands/init.ts:150` and shares the same flaw: by the time a user reaches this code path their machine is already in `recipients` from `init`, so the suggested command will fail on the existing recipient's machine with `"Recipient '<name>' already exists"` (`src/commands/key.ts:90-96`). The `pull.test.ts` coverage only asserts the message string is built — it doesn't simulate the end-to-end remediation flow, so the broken hint passes CI.

**Suggested fix:** Resolve once at the source (e.g. make `key add` idempotent when the pubkey matches, or stop having `init` push the recipient itself), then keep this hint pointing at the now-working command. Either way, the `pull.test.ts` recipient-handoff describe block should also exercise the full `init → pull (fail) → key add (recovers) → pull (success)` sequence so a future regression is caught.

src/commands/__tests__/integration.test.ts
```diff
@@ -261,6 +261,51 @@ describe("integration", () => {
...
+    // The init flow should warn the new joiner that they cannot pull until an
+    // existing recipient runs `key add` for them. The message must include
+    // both the local machine name and its pubkey so it is copy-pasteable.
+    const hint = fakeLogs.warn.find((message) =>
+      message.includes("agentsync key add machine-beta"),
+    );
+    expect(hint).toBeDefined();
+    expect(hint).toContain(machineB.recipient);
+    expect(hint).toContain("`agentsync pull` on this machine will fail");
+  });
```
chrisleekr, line 290: _💭 Suggestion_      | _🟡 Minor_

**Test asserts the hint text but not that the hint actually works.**

This test only verifies the warn message is emitted. It doesn't execute `agentsync key add machine-beta <pubkey>` on `machineA` to confirm the suggested remediation actually decrypts the vault for `machineB`. Because of the duplicate-recipient guard in `src/commands/key.ts:90-96`, that command currently fails — but no test catches it, so the broken hint shipped clean through CI.

**Suggested fix:** Extend this test (or add a sibling) that runs `init` on machine-beta, then runs `keyMod.keyCommand.subCommands.add.run({ name: 'machine-beta', pubkey: machineB.recipient })` on machine-alpha, then drives `performPull` on machine-beta and asserts it succeeds. That tests the documented onboarding flow end-to-end and prevents future regressions.

chrisleekr:

## Outstanding

Nothing blocking merge. CI green; all three review threads resolved. Note: PR reviewDecision is not APPROVED (self-review only, no formal approval was given), so a human approval is still required before merge per the project's "merging is a human action" rule.

_cost: $4.0871 · turns: 50 · duration: 430s_

src/commands/init.ts
chrisleekr: _✅ Addressed_, commit `6eda2ea`

**Made `key add` idempotent when `recipients[name] === pubkey` so the onboarding hint actually works.**

You were right — by the time this warn fires, `init` has already pushed the joining machine into `recipients`, so the suggested `agentsync key add` dead-ended at the duplicate-name guard (`src/commands/key.ts:75-79` and `90-101`). Fixed both checks at `src/commands/key.ts:75` and `src/commands/key.ts:95` to compare pubkey on a name collision: same pubkey → proceed to re-encrypt the vault (the actual purpose of the step); different pubkey → still error so the existing collision guard exercised by the "key add re-checks aliases after reconciling with a newer remote config" test stays intact. Hint at `src/commands/init.ts:150` is now the working command.

src/commands/pull.ts
chrisleekr: _✅ Addressed_, commit `6eda2ea`

**Same root fix in `src/commands/key.ts` unblocks this hint too.**

The hint at `src/commands/pull.ts:105` was the same broken `agentsync key add` command. With `key add` now idempotent on matching pubkey (`src/commands/key.ts:75`, `src/commands/key.ts:95`), the suggested remediation succeeds and re-encrypts the vault for the joining machine. The new integration test `recipient handoff: init + key add (idempotent on same pubkey) re-encrypts vault for new machine` in `src/commands/__tests__/integration.test.ts` drives the full init → key add → decrypt sequence end-to-end so a regression on either side fails CI.

src/commands/__tests__/integration.test.ts
chrisleekr: _✅ Addressed_, commit `6eda2ea`

**Added the end-to-end test that exercises the suggested `key add` against a real vault.**

The new test in `src/commands/__tests__/integration.test.ts` ("recipient handoff: init + key add (idempotent on same pubkey) re-encrypts vault for new machine") seeds an `.age` artifact encrypted only for machine-alpha, runs `init` on machine-beta (which now writes machine-beta into `recipients` on the remote), then runs `keyMod.keyCommand.subCommands.add` with `name: "machine-beta", pubkey: machineB.recipient` on machine-alpha. It asserts the command succeeds, then pulls on machine-beta and decrypts the re-encrypted artifact with `machineB.identity`. Future regressions of either the `init`-side write or the `key add` idempotency will fail this test.

chrisleekr:

🔍 **Code review complete**, 10 files, +412/-30.

## Summary

The PR delivers both UX papercuts called out in #44: ENOENT before `init` is now a single-line friendly error via the new `loadVaultConfigOrExit` helper, and second-machine `pull` translates the raw age `no identity matched any of the file's recipients` error into a copy-pasteable `agentsync key add` hint. The previous review thread caught the broken-hint regression (the suggested `key add` would dead-end at the duplicate-name guard); commit `6eda2ea` fixed it by making `key add` idempotent when `recipients[name] === pubkey`, and the new end-to-end integration test exercises the full `init → key add → pull (decrypts)` sequence on a real bare repo. The remaining concern is narrow: in `pull.ts`, lifting `resolveRuntimeContext` and `loadVaultConfigOrExit` out of the outer try/catch narrows the documented "never throws" contract for non-ENOENT failures (zod schema errors, mkdir failures), which would surface as a Node stack trace via the CLI path — undermining the PR's "friendly errors" goal in a related-but-not-fixed corner. One Minor inline finding posted; no Blockers or Majors.

## What was checked

Files read in full:
- `src/commands/shared.ts` — new `loadVaultConfigOrExit` helper, ENOENT branch + re-throw semantics
- `src/commands/init.ts` — `joinedExistingVault` derivation + post-init warn placement
- `src/commands/key.ts` — both pre-reconcile and post-reconcile name-collision guards, idempotency on matching pubkey
- `src/commands/pull.ts` — extracted runtime/config load, per-agent try/catch around `apply`, `AGE_NO_IDENTITY_MARKER` detection, `buildRecipientHandoffHint`
- `src/commands/push.ts`, `src/commands/status.ts`, `src/commands/skill.ts` — helper migration
- `src/commands/__tests__/integration.test.ts` — recipient-handoff hint test, end-to-end re-encryption test, first-machine negative test
- `src/commands/__tests__/pull.test.ts` — real-vault fixture seeding, marker-translation + non-age propagation tests
- `src/commands/__tests__/shared.test.ts` — ENOENT exit and non-ENOENT re-throw coverage

Cross-references performed:
- `grep loadConfig(resolveConfigPath` — confirmed `src/daemon/index.ts:61,71` and `src/commands/doctor.ts:174` deliberately keep the raw `loadConfig` (both have context-specific error handling: daemon's "Startup failed:" + exit; doctor's failed-check row). Skipping the migration there is correct.
- `grep loadVaultConfigOrExit` — five command sites migrated; no orphan usages
- `git show origin/main:src/commands/pull.ts` — confirmed the pre-PR `loadConfig` + `resolveRuntimeContext` calls were both INSIDE the outer try/catch; the new code lifts them out, which is the source of the contract narrowing flagged
- `git show origin/main:src/commands/push.ts` and `skill.ts` and `status.ts` — confirmed `loadConfig` was already outside any try/catch in those files; the helper migration is behavior-preserving there
- `src/config/loader.ts` — `loadConfig` raises ENOENT from `readFile` or zod parse errors for malformed TOML; the ENOENT discriminator in shared.ts is therefore tight
- `src/core/encryptor.ts:51-53` — `identityToRecipient` wraps `age.identityToRecipient` directly; the try/catch fallback inside `buildRecipientHandoffHint` is defensive against a malformed identity string (very rare since `loadPrivateKey` already succeeded earlier in `performPull`)
- `src/commands/__tests__/integration.test.ts:525` — "key add re-checks aliases after reconciling with a newer remote config" still verifies the genuine-conflict case (different pubkey on existing name) errors out, so the idempotency relaxation hasn't weakened the collision guard

Validation runs:
- `bun run typecheck` — clean
- `bun run lint` — 8 warnings, all in `docs/stylesheets/extra.css` (unrelated to this PR's files)
- `bun test src/commands/__tests__/shared.test.ts src/commands/__tests__/pull.test.ts` — 10/10 pass, 0 fail
- `bun test src/commands/__tests__/integration.test.ts` — 22/22 pass including the three new recipient-handoff tests

## Findings

[minor] `src/commands/pull.ts:44-45` — `resolveRuntimeContext()` and `loadVaultConfigOrExit(...)` were moved outside the outer try/catch in `performPull`. Pre-PR, schema/parse errors from `loadConfig` (zod) and `mkdir` failures inside `resolveRuntimeContext` were caught and surfaced as `result.errors[0]` with `fatal: true`. Now they throw straight out of `performPull`; the CLI wrapper at `src/commands/pull.ts:125-141` does not catch, so a malformed `agentsync.toml` would crash with a Node stack trace — exactly the failure-mode shape this PR removes for the sibling ENOENT case. The lift was unnecessary for the ENOENT path because `process.exit(1)` terminates the process regardless of try/catch placement. **Recommended fix:** move both calls back inside the outer try block (or wrap them in their own try/catch that pushes to `errors`); ENOENT still exits cleanly, non-ENOENT regains the pre-PR "friendly error row" semantics so the CLI logs via `log.error` instead of crashing.

## Reasoning

**Why I didn't escalate the lift-out-of-try as Major.** The non-ENOENT failure modes that would hit this path require a user to have manually corrupted `agentsync.toml` or to run on a filesystem where `~/.agentsync` cannot be created. Both are rare. The user-visible difference is "wrapped zod error message in `result.errors`" vs "raw stack trace via citty's default handler" — both communicate roughly the same information, and either way the user knows to inspect their TOML. The fix is trivially safe though, and the PR's stated goal explicitly is to remove this exact UX shape, so I flagged it as Minor for completeness.

**Why the idempotency fix in `key.ts:75` and `key.ts:95` is the right shape.** The original review correctly identified that `init` writes the joining machine's pubkey into `recipients` before exiting, so by the time `key add <machine> <pubkey>` runs on the existing recipient, `recipients[machine]` already equals `pubkey`. The fix relaxes the duplicate-name guard *only* when the existing entry's pubkey matches — different pubkey on same name still errors, which keeps the pre-existing "you cannot silently overwrite a different recipient" invariant intact. Verified by inspection: the test at `integration.test.ts:525` ("key add re-checks aliases after reconciling with a newer remote config") still drives the genuine-conflict case and asserts `process.exitCode === 1` plus the "already exists" error message, so the relaxation hasn't weakened the genuine-collision path.

**Why the new end-to-end test at `integration.test.ts:292` is load-bearing.** It seeds an `.age` artifact encrypted only for `machine-alpha`, runs `init` on `machine-beta` (which pushes its pubkey into `recipients`), then runs the exact `key add machine-beta <pubkey>` the new hint suggests on `machine-alpha`, and finally decrypts the re-encrypted artifact on `machine-beta` with its own identity. Any future regression that either (a) breaks `init` writing recipients itself, or (b) reintroduces the non-idempotent duplicate-name guard, would fail this test on the `expect(decrypted).toBe(plaintext)` assertion at line 355.

**Why I didn't flag the substring-match for the age marker.** `AGE_NO_IDENTITY_MARKER = "no identity matched any of the file's recipients"` is brittle to upstream wording changes, but the PR description calls this out as a known risk and the test at `pull.test.ts:102-128` throws the literal marker string from a fake agent, so the next age-library bump that reworded the error message would fail this test loudly. The alternative (matching by a subclassed error type) requires upstream changes. Acceptable.

**Why I didn't flag daemon/doctor not being migrated.** Both `src/daemon/index.ts:61,71` and `src/commands/doctor.ts:174` keep the raw `loadConfig` because each has context-specific error handling: the daemon wraps with "Startup failed: …" + `process.exit(1)`, and `doctor` reports failed checks as a row in its diagnostic table. Migrating them to `loadVaultConfigOrExit` would replace those context-specific UX paths with the generic "vault not initialized" message, which would be wrong for both. Correct call by the author.

**Why I didn't flag the per-agent try/catch in `pull.ts:67-79`.** Compared to the pre-PR single-outer-try structure, the new layout preserves non-AGE error semantics: a non-marker `throw err` re-throws to the outer catch and ends up in `errors[]`, same as before. The early `return` on the AGE marker is correct because all agents share the same recipient set — if one fails decryption, every other agent will too, and the hint already covers the remediation. One subtle property: if multiple agents would each fail with non-AGE errors, only the first surfaces — but that was also the pre-PR behaviour (single outer try, first throw wins), so no regression.

**Why the `buildRecipientHandoffHint` `identityToRecipient` fallback is fine.** The fallback message at `pull.ts:95` ("unable to derive local pubkey — re-run after `agentsync init`") is essentially defensive: by the time we reach this branch, `loadPrivateKey` already succeeded inside the try block, so the identity string is well-formed. `age.identityToRecipient` would only throw on a malformed bech32 identity, which would also have prevented the failing `agent.apply` call from reaching the AGE marker path in the first place. The fallback is dead code in practice, but not worth flagging.

_cost: $4.0547 · turns: 42 · duration: 466s_

src/commands/pull.ts
```diff
@@ -39,9 +41,9 @@ export async function performPull(
   const errors: string[] = [];
   let applied = 0;
   let fatal = false;
+  const runtime = await resolveRuntimeContext();
+  const config = await loadVaultConfigOrExit(runtime.vaultDir);
```
chrisleekr: _💭 Suggestion_      | _🟡 Minor_

**Lifting `resolveRuntimeContext` + `loadVaultConfigOrExit` out of the try block narrows the documented "never throws" contract for non-ENOENT failures.**

Pre-PR, `loadConfig(...)` and `resolveRuntimeContext()` sat inside the outer try at the old `pull.ts` and any schema/parse failure (zod), or an `mkdir` failure inside `resolveRuntimeContext`, became `result.errors[0]` with `fatal: true`. With both calls moved outside the try block at `src/commands/pull.ts:44-45`, only the ENOENT path is friendly (handled by `loadVaultConfigOrExit`'s `process.exit(1)`); every other failure now throws straight out of `performPull`. The CLI wrapper at `src/commands/pull.ts:125-141` does not catch the throw, so a user-edited but malformed `agentsync.toml` would now surface as a Node stack trace via citty's default handler — exactly the kind of UX this PR sets out to remove for the sibling ENOENT case. The test comment at `src/commands/__tests__/pull.test.ts:54-55` ("performPull never throws when called against a real vault") is now narrower than the new behavior in a way no test asserts. Note: `process.exit(1)` terminates the process regardless of try/catch placement, so lifting the call out of the try block was not required to make the missing-vault path work — it only widens the exposure of other failure modes.

**Suggested fix:** Move both `resolveRuntimeContext()` and `loadVaultConfigOrExit(...)` back inside the outer `try { ... } catch` (or wrap the calls in their own try/catch that pushes to `errors`). The ENOENT branch still exits cleanly via `process.exit(1)`; non-ENOENT failures regain their pre-PR "friendly error row" semantics so the CLI logs them through `log.error` instead of crashing with a stack trace.

chrisleekr:

## Outstanding

Nothing blocks merge mechanically: branch is `MERGEABLE`, `mergeStateStatus` is `CLEAN`, CI is green, every review thread is resolved. The only gating item is `reviewDecision` being empty — i.e. a human approval is still required per the project's "merging is a human action" rule, which is expected.

_cost: $1.8971 · turns: 30 · duration: 471s_

src/commands/pull.ts
chrisleekr: _✅ Addressed_, commit `3f3abb9`

**Moved `resolveRuntimeContext` and `loadVaultConfigOrExit` back inside the outer try block.**

You're right: lifting them out narrowed the friendly-error contract for non-ENOENT failures (zod schema errors, mkdir EACCES) so they would have surfaced as a citty-default Node stack trace. The fix at `src/commands/pull.ts:50-53` puts both calls back inside the outer try so those now become an `errors[]` row again. The ENOENT branch is unaffected — `loadVaultConfigOrExit`'s `process.exit(1)` terminates before any catch runs. Test comment at `src/commands/__tests__/pull.test.ts:53-57` updated to match.

chrisleekr:

## Outstanding

Nothing mechanically blocks merge: branch is `MERGEABLE`, `mergeStateStatus` is `CLEAN`, every check passes, every review thread on the PR is resolved. The only gating item is `reviewDecision` being empty — a human APPROVE is still required per the project's "merging is a human action" rule. The bot will not call `gh pr merge`.

_cost: $3.1653 · turns: 5 · duration: 47s_

src/commands/__tests__/integration.test.ts
```diff
@@ -261,6 +261,117 @@ describe("integration", () => {
...
+    await withMachineEnv(firstMachine, async () => {
+      await initMod.initCommand.run?.({
+        args: { remote: bareRepoPath, branch: "main" },
+        rawArgs: [],
+        cmd: {} as never,
+      } as never);
+    });
+
+    expect(fakeLogs.warn.some((message) => message.includes("agentsync key add"))).toBe(false);
+  });
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `00c7ebe`

**Pinned the negative assertion to the canonical lead phrase of the init hint.**

Added a second assertion at `src/commands/__tests__/integration.test.ts` that checks `fakeLogs.warn` does not contain `"This machine is registered but cannot decrypt"` (the canonical lead phrase from `src/commands/init.ts:148`), kept alongside the existing `"agentsync key add"` check. Any future rewording (e.g. `key set`, `recipient add`) still has to clear both guards before it can leak onto the first-machine bootstrap path.

src/commands/__tests__/pull.test.ts
```diff
@@ -22,3 +65,82 @@ describe("performPull", () => {
...
+      const value = savedEnv[key];
+      if (value === undefined) delete process.env[key];
+      else process.env[key] = value;
+    }
+    await rm(tmpDir, { recursive: true, force: true });
+  });
+
+  afterAll(() => {
+    mock.restore();
+  });
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `00c7ebe`

**Added an outer-try regression test that corrupts `agentsync.toml` and asserts the friendly `errors[]` row.**

New test in `src/commands/__tests__/pull.test.ts` writes garbage TOML into the seeded vault, runs `performPull`, and asserts `result.fatal === true` with `errors.length === 1`. This exercises the path you flagged: `loadVaultConfigOrExit` re-throws (non-ENOENT), the outer try in `performPull` catches it, and the failure surfaces as an `errors[]` row instead of a raw Node throw. I chose TOML parse failure over a read-only vault dir because the dir is created inside `resolveAgentSyncHome` which the test fixture already controls via env vars; permission tricks would have been more fragile across CI runners.

src/commands/__tests__/pull.test.ts
```diff
@@ -22,3 +65,82 @@ describe("performPull", () => {
...
+          throw new Error("disk full");
+        },
+      },
+    ]);
+
+    const result = await performPull({});
+    expect(result.fatal).toBe(true);
+    expect(result.errors[0]).toContain("disk full");
+    expect(result.errors[0] ?? "").not.toContain("agentsync key add");
+  });
```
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `00c7ebe`

**Dropped the `?? ""` fallback on the negative assertion.**

`src/commands/__tests__/pull.test.ts` non-age propagation test now reads `expect(result.errors[0]).not.toContain("agentsync key add")` so a regression that produces an empty `errors[]` would fail loudly with "Cannot read properties of undefined" instead of silently passing. Symmetric with the positive line above.

PR #61, opened 2026-05-11
test(watcher): use polling waitFor for positive assertions

### LINKED ISSUE #45: Watcher tests are timing-flaky under full-suite load
## Summary

The `Watcher > callback fires exactly once for rapid writes within the debounce window` test failed once during a full 405-test run on macOS arm64 but passed 3/3 times when the file was run in isolation. The other two watcher tests (`Watcher.remove`, `Watcher.close`) use the same sleep-and-assert pattern with shorter buffers (50ms / 150ms) and are likely fragile under the same conditions.

## Failure observed

```
src/core/__tests__/watcher.test.ts:
46 |     // Debounce must collapse all writes into exactly one callback
47 |     expect(fired.length).toBe(1);
                              ^
error: expect(received).toBe(expected)
Expected: 1
Received: 0

(fail) Watcher > callback fires exactly once for rapid writes within the debounce window [256.95ms]
```

Three subsequent isolated runs passed:

```
--- run 1 ---
(pass) Watcher > callback fires exactly once for rapid writes within the debounce window [257.58ms]
--- run 2 ---
(pass) Watcher > callback fires exactly once for rapid writes within the debounce window [257.65ms]
--- run 3 ---
(pass) Watcher > callback fires exactly once for rapid writes within the debounce window [258.75ms]
```

## Root cause

`fs.watch` on macOS (FSEvents) batches events with non-deterministic latency. The test writes 5 files inside a 100ms debounce window, then sleeps 250ms. Under full-suite I/O contention, the FSEvents callback can land >150ms after the final write, by which time the assertion runs but the debounce timer has not yet fired.

## Proposed fix

Replace `await Bun.sleep(N); expect(fired.length).toBe(1)` with a polling helper:

```ts
async function waitFor(condition: () => boolean, maxMs = 2000, stepMs = 25) {
  const start = Date.now();
  while (!condition()) {
    if (Date.now() - start > maxMs) throw new Error("timeout");
    await Bun.sleep(stepMs);
  }
}

// in the debounce test
for (let i = 0; i < 5; i++) await writeFile(filePath, `write-${i}`, "utf8");
await waitFor(() => fired.length >= 1);
await Bun.sleep(150);  // give any spurious extras a chance
watcher.close();
expect(fired.length).toBe(1);
```

Apply the same pattern to:

- `Watcher.remove stops callbacks…` (currently 150ms sleep)
- `Watcher.close stops all watchers…` (currently 150ms sleep)

## Acceptance

- [ ] All three watcher tests pass deterministically when invoked as part of `bun test` 5+ consecutive times.
- [ ] No fixed `Bun.sleep` followed by a positive assertion (the negative assertions in `remove` and `close` can keep a sleep, but with a longer margin).

chrisleekr:
## Summary

Watcher tests on macOS (FSEvents) fail intermittently under full-suite I/O contention because the previous tests awaited the debounce window with a fixed `Bun.sleep` before positive equality assertions — the FS event can land later than the sleep budget, and the assertion fires while `fired.length === 0`. This PR replaces fixed-sleep-before-positive-assert with a local polling helper, keeps a (widened) fixed sleep before the two negative assertions, and adds a warm-up handshake to the close test so the watcher is provably live before close.

The fix is test-only; `src/core/watcher.ts` is unchanged.

## Changes

- Added a local `waitFor(condition, maxMs=2000, stepMs=25)` poll helper to `src/core/__tests__/watcher.test.ts` that throws a descriptive timeout error so failures surface as `waitFor: condition not met` rather than `Expected 1, Received 0`.
- Rewrote the debounce test to `waitFor(() => fired.length >= 1)`, then sleep 150 ms for spurious extras, then close and assert exactly one callback.
- Rewrote `Watcher.remove` to await the first callback via `waitFor` before snapshotting `beforeRemove`. Negative assertion keeps a fixed sleep but widens it from 150 ms to 300 ms.
- Rewrote `Watcher.close` to replace the 50 ms warm-up sleep with a warm-up write + `waitFor` so the watcher is provably live before close. Negative-assertion sleep widened to 300 ms; the post-close baseline is now `beforeClose` rather than a hard-coded `0`, so any warm-up event counts as part of the pre-close baseline.

### Architecture

Test-only change; no production behaviour or flow shift.

## Test Evidence

- `for i in 1 2 3 4 5; do bun test src/core/__tests__/watcher.test.ts; done` — 5/5 runs pass, `3 pass / 0 fail / 3 expect() calls` each.
- `bun test` (full suite, ×2) — `239 pass / 20 fail / 16 errors`, identical to baseline on `main` (the 20 failures and 16 errors are pre-existing test bleed unrelated to this change; confirmed by checking out main and re-running).
- `bun run lint` — only pre-existing CSS warnings on `docs/...` (mkdocs site); no new findings from this change.

## Files changed

- `src/core/__tests__/watcher.test.ts` — added `waitFor` helper; rewrote three tests to await callback arrival (positive) or sleep + widened margin (negative).

## Commits

- `39c28ce` — test(watcher): use polling waitFor for positive assertions

## Tests run

- `bun test src/core/__tests__/watcher.test.ts` (×5) · 3/3 pass per run
- `bun test` (×2) · 239 pass / 20 fail / 16 errors (baseline-matching pre-existing failures)
- `bun run typecheck` · pass
- `bun run lint` · 8 pre-existing CSS warnings, no errors

## Verification

1. Watcher suite passes deterministically in isolation (5+ consecutive `bun test src/core/__tests__/watcher.test.ts`).
2. Full `bun test` runs show no Watcher failures across two consecutive runs.
3. Diff inspection:
   - No `Bun.sleep(...)` immediately precedes a positive `toBe(1)`-style assertion in `watcher.test.ts`.
   - Both negative assertions (`Watcher.remove`, `Watcher.close`) retain a fixed sleep but with a wider margin (300 ms vs original 150 ms).
   - No production code in `src/core/watcher.ts` was modified.

## Risks / Follow-ups

- The fix targets macOS FSEvents latency, which cannot be reproduced on the Linux CI host. CI cannot exercise the failure mode directly; the fix is structural (await the event vs assume a budget) and the worst-case effect on Linux is a per-test slowdown bounded by `maxMs = 2000`.
- The 20 pre-existing full-suite failures (mock.module bleed of `node:fs/promises`) are out of scope and not addressed here.

Closes #45

## Summary by CodeRabbit

* **Tests**
  * Improved file system watcher test reliability and stability across platforms by replacing fixed timing waits with event-driven polling conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/61)

### DIFF
--- a/src/core/__tests__/watcher.test.ts
+++ b/src/core/__tests__/watcher.test.ts
@@ -55,8 +55,11 @@ describe("Watcher", () => {
     // Await the actual callback rather than a fixed sleep — FSEvents latency
     // can exceed any fixed budget under full-suite I/O contention.
     await waitFor(() => fired.length >= 1);
-    // Give any spurious extras a chance to land before snapshotting.
-    await Bun.sleep(150);
+    // Watch for a spurious extra. Inverting the poll makes the "no extras
+    // within window" intent explicit and matches the 300 ms margin used by
+    // the other negative branches — a fixed sleep here would be silently
+    // masked by close() cancelling any still-pending debounce timer.
+    await waitFor(() => fired.length > 1, 300).catch(() => {});
     watcher.close();
 
     // Debounce must collapse all writes into exactly one callback
--- a/src/core/__tests__/watcher.test.ts
+++ b/src/core/__tests__/watcher.test.ts
@@ -13,6 +13,19 @@ import { Watcher } from "../watcher";
   mock.module("node:fs/promises", () => realFsPromises);
 }
 
+// Poll for an event-driven condition instead of a fixed sleep — fs.watch on
+// macOS (FSEvents) batches with non-deterministic latency, so positive
+// assertions need to await callback arrival rather than assume a budget.
+async function waitFor(condition: () => boolean, maxMs = 2000, stepMs = 25): Promise<void> {
+  const start = Date.now();
+  while (!condition()) {
+    if (Date.now() - start > maxMs) {
+      throw new Error(`waitFor: condition not met within ${maxMs}ms`);
+    }
+    await Bun.sleep(stepMs);
+  }
+}
+
 // Watcher debounce and lifecycle
 
 describe("Watcher", () => {
@@ -39,8 +52,11 @@ describe("Watcher", () => {
       await writeFile(filePath, `write-${i}`, "utf8");
     }
 
-    // Wait debounce window + generous buffer
-    await Bun.sleep(250);
+    // Await the actual callback rather than a fixed sleep — FSEvents latency
+    // can exceed any fixed budget under full-suite I/O contention.
+    await waitFor(() => fired.length >= 1);
+    // Give any spurious extras a chance to land before snapshotting.
+    await Bun.sleep(150);
     watcher.close();
 
     // Debounce must collapse all writes into exactly one callback
@@ -57,16 +73,18 @@ describe("Watcher", () => {
       fireCount++;
     });
 
-    // First write — should trigger
+    // First write — must observe the callback before snapshotting, otherwise
+    // the negative assertion can pass for the wrong reason on slow event delivery.
     await writeFile(filePath, "initial", "utf8");
-    await Bun.sleep(150);
+    await waitFor(() => fireCount >= 1);
     const beforeRemove = fireCount;
 
     watcher.remove(tmpDir);
 
-    // Write after remove — must NOT trigger
+    // Write after remove — must NOT trigger. Use a wider margin than the
+    // original 150 ms so FSEvents latency can't masquerade as a quiet watcher.
     await writeFile(filePath, "after-remove", "utf8");
-    await Bun.sleep(150);
+    await Bun.sleep(300);
 
     expect(fireCount).toBe(beforeRemove); // no new events
   });
@@ -75,18 +93,24 @@ describe("Watcher", () => {
   test("Watcher.close stops all watchers; writes after close invoke no callbacks", async () => {
     const watcher = new Watcher();
     let fireCount = 0;
+    const warmupPath = join(tmpDir, "warmup.txt");
     const filePath = join(tmpDir, "post-close.txt");
 
     watcher.add(tmpDir, 50, () => {
       fireCount++;
     });
-    await Bun.sleep(50); // let the watcher initialise
+
+    // Warm-up write proves the watcher is live before we close it; without
+    // this, a slow FSEvents subscribe could let a pre-close event leak past.
+    await writeFile(warmupPath, "warmup", "utf8");
+    await waitFor(() => fireCount >= 1);
+    const beforeClose = fireCount;
 
     watcher.close();
 
     await writeFile(filePath, "after-close", "utf8");
-    await Bun.sleep(150);
+    await Bun.sleep(300);
 
-    expect(fireCount).toBe(0);
+    expect(fireCount).toBe(beforeClose);
   });
 });


chrisleekr:

🔍 **Code review complete**, 1 files, +33/-9.

## Summary

Test-only flake fix for `src/core/__tests__/watcher.test.ts`: replaces fixed-sleep-before-positive-assert with a local `waitFor` polling helper, widens the negative-assertion sleeps from 150 ms to 300 ms, and adds a warm-up handshake to the `Watcher.close` test so the watcher is provably live before close. Production `src/core/watcher.ts` is untouched. The change is well-motivated, the trade-offs are documented in the PR description, and locally I get 3/3 pass × 3 runs, clean typecheck, and lint warnings limited to pre-existing docs/CSS. One **minor** observation about a residual fixed-budget assumption in the debounce test; otherwise LGTM.

## What was checked

- Read `src/core/__tests__/watcher.test.ts` in full (the only file changed, +33/-9).
- Read `src/core/watcher.ts` in full to confirm test semantics match production behaviour — especially `Watcher.remove`'s synchronous `FSWatcher.close()` + `clearTimeout` pair (`src/core/watcher.ts:59-65`) which is load-bearing for the negative assertions.
- Searched repo-wide for other consumers of timing-sensitive sleeps (`grep -rn "Bun.sleep" src/ --include="*.test.ts"`) — no other tests share this pattern, so the `waitFor` helper does not need to be hoisted to `src/test-helpers/fixtures.ts` for reuse right now.
- Re-ran `bun test src/core/__tests__/watcher.test.ts` ×3 — 3/3 passes per run.
- `bun run lint` — only the pre-existing 8 CSS warnings on docs files, no new findings from this diff.
- Verified the diff does not touch `src/core/watcher.ts`, `src/daemon/`, or anything outside the test file.

## Findings

[minor] `src/core/__tests__/watcher.test.ts:60` — Spurious-extras window stays at 150 ms while the negative-assertion sleeps in `Watcher.remove`/`Watcher.close` were widened to 300 ms. Because `watcher.close()` runs **before** the `expect`, a delayed FSEvents batch whose debounce timer is still pending at close time will be cancelled by `Watcher.remove` (`src/core/watcher.ts:63`) and silently masked — exactly the failure mode the rest of the PR is hardening against. Recommend either widening to 300 ms to match, or inverting with `waitFor(() => fired.length > 1, 300).catch(() => {})` so the "no extras within window" intent is explicit and not load-bearing on `close()` cancelling pending timers. Inline comment posted.

## Reasoning

- **`waitFor` helper (lines 19-27).** Correct: condition-first, time-check, sleep loop; descriptive throw on timeout. The throw text (`waitFor: condition not met within {maxMs}ms`) is what was promised in the PR description and what a failing CI run will surface instead of `Expected 1, Received 0`. Default `maxMs = 2000` bounds the slowdown on Linux to a per-test worst case of 2s, which matches the PR's stated risk envelope. Not flagged.

- **Test 1 (debounce, lines 43-64).** `waitFor(() => fired.length >= 1)` correctly awaits arrival of the first callback. The follow-up `Bun.sleep(150)` plus the subsequent `expect(fired.length).toBe(1)` is the part I flagged above — see Findings. The `watcher.close()` immediately before the assertion is sound and does not itself leak (`close()` calls `remove()` for each path which cancels pending timers).

- **Test 2 (Watcher.remove, lines 67-90).** The pattern is: confirm first callback, snapshot, remove, write, sleep 300 ms, assert no delta. The 300 ms widening is the right fix. There is no explicit `watcher.close()` at the end, but `Watcher.remove(tmpDir)` is the only watched path's removal, which is equivalent — and `afterEach` removes the whole `tmpDir`. The race I checked: between `waitFor` returning and `watcher.remove(tmpDir)`, a second FSEvents batch could set a new debounce timer; `remove()` cancels it via `clearTimeout` at `src/core/watcher.ts:63`. Safe.

- **Test 3 (Watcher.close, lines 93-115).** Replacing the unconditional `Bun.sleep(50)` warm-up with `writeFile(warmupPath) + waitFor(() => fireCount >= 1)` is the right structural change — the original had a false-pass mode where if the watcher had not yet subscribed when `close()` ran, the assertion `fireCount === 0` would have been trivially true. Using `beforeClose = fireCount` (instead of a hard-coded `0`) correctly admits any spurious warm-up events as part of the pre-close baseline. Two different file names for warm-up vs post-close write (`warmup.txt` vs `post-close.txt`) is defensive against any FS-level event coalescing on identical writes; no bug.

- **Cross-cutting checks.** No production code changed, so no API/contract impact. No new lint or typecheck findings. The `waitFor` helper is private to the file and not duplicated elsewhere; hoisting it to test-helpers would be premature given a single consumer. The fix targets macOS FSEvents specifically and the failure mode cannot be reproduced on the Linux CI host — the PR description correctly acknowledges this and accepts that CI proves only "no regression on Linux" not "fixes the macOS flake".

_cost: $1.9750 · turns: 23 · duration: 293s_

src/core/__tests__/watcher.test.ts
```diff
@@ -39,8 +52,11 @@ describe("Watcher", () => {
...
     }
 
-    // Wait debounce window + generous buffer
-    await Bun.sleep(250);
+    // Await the actual callback rather than a fixed sleep — FSEvents latency
+    // can exceed any fixed budget under full-suite I/O contention.
+    await waitFor(() => fired.length >= 1);
+    // Give any spurious extras a chance to land before snapshotting.
+    await Bun.sleep(150);
     watcher.close();
```
chrisleekr, line 63: _💭 Suggestion_      | _🟡 Minor_

**Spurious-extras window is still a fixed budget that `watcher.close()` can silently mask.**

At `src/core/__tests__/watcher.test.ts:57-63`, after `waitFor(() => fired.length >= 1)` the test sleeps 150 ms then calls `watcher.close()` before asserting `fired.length === 1`. If FSEvents delivers a second batch >150 ms after the first (the same kind of late delivery this PR is trying to defend against), the second debounce timer will be **pending** when `close()` runs and gets cancelled by `Watcher.remove()` at `src/core/watcher.ts:63` — the assertion then passes even though a real debounce-leak bug would have produced a second callback. The negative branch on Watcher.remove/close was widened to 300 ms for the same reason; this site kept the 150 ms.

**Suggested fix:** Either widen this sleep to match the 300 ms used in the other negative branches, or invert the check with `waitFor(() => fired.length > 1, 300).catch(() => {})` so the intent ("no extras within window") is explicit and not dependent on the subsequent `close()` cancelling pending timers.

chrisleekr:

## Outstanding

Nothing blocks merge. CI green, the one open thread is resolved, no other review comments outstanding. `reviewDecision` is empty (no human reviewer approval requested), so a maintainer needs to merge manually per the no-bot-merge rule.

_cost: $1.5933 · turns: 5 · duration: 39s_

src/core/__tests__/watcher.test.ts
chrisleekr: _✅ Addressed_, commit `dfdf492`

**Inverted the spurious-extras poll so the assertion no longer relies on `close()` cancelling pending timers.**

Replaced the 150 ms fixed sleep at `src/core/__tests__/watcher.test.ts:59` with `await waitFor(() => fired.length > 1, 300).catch(() => {})`. The 300 ms now matches the other negative branches, the intent ("no extras within window") is explicit, and a delayed FSEvents batch arriving during the window will be observed by the poll instead of being silently cancelled by `Watcher.remove()` at `src/core/watcher.ts:63`. Verified: `bun test src/core/__tests__/watcher.test.ts` 3/3 pass × 3 runs; typecheck clean; no new lint findings.

PR #64, opened 2026-05-12
refactor(init): remove dead readFile after writeConfig

### LINKED ISSUE #48: init.ts: dead readFile call after writeConfig
## Summary

`src/commands/init.ts:128` performs an `await readFile(configPath, "utf8")` whose result is discarded. It runs immediately after `writeConfig(configPath, config)` writes the same file. Looks like a leftover from a verify-after-write that was never wired up.

## Trace

```ts
// src/commands/init.ts (around line 124-128)
await writeConfig(configPath, config);

const gitignorePath = join(runtime.vaultDir, ".gitignore");
await writeFile(gitignorePath, "*.tmp\n", "utf8");
await readFile(configPath, "utf8");   // ← result not assigned, not used
```

## Proposed fix

Delete the line. If the original intent was a post-write integrity check, replace it with a `loadConfig(configPath)` call (which already round-trips through Zod). Otherwise the file's existence is already implied by `writeConfig` resolving without error.

## Acceptance

- [ ] Line removed (or replaced with a meaningful round-trip check).
- [ ] All init tests still pass.

chrisleekr:
## Summary

Removes a dead `await readFile(configPath, "utf8")` in `src/commands/init.ts` whose return value was discarded — a leftover from a verify-after-write that was never wired up. `writeConfig` already round-trips through TOML and `loadConfig` validates via Zod, so file existence/validity is implied by `writeConfig` resolving without error; no replacement check is needed. Also drops the now-unused `readFile` import.

Closes #48

## Changes

- Remove the dead `await readFile(configPath, "utf8")` call that sat between the `.gitignore` write and `git.commit` in the init flow.
- Trim `readFile` from the `node:fs/promises` import; nothing else in `init.ts` references it.

## Files changed

- `src/commands/init.ts` · remove discarded `await readFile(configPath, "utf8")` and drop the now-unused `readFile` import.

## Commits

- `c502093` · refactor(init): remove dead readFile after writeConfig

## Tests run

- `bun run typecheck` · pass (no errors).
- `bun run lint` · pass (8 pre-existing CSS warnings + 1 info on unrelated files; no errors).
- `bun test src/commands/__tests__/integration.test.ts` · pass (27 pass, 0 fail) — covers fresh init, join-existing-vault, and error-rollback paths through `initMod.initCommand.run`.
- `bun test` (full) · 239 pass / 20 fail / 16 errors — same totals on `main` at `c5642a6` (pre-existing test infra failures in `daemon.test.ts`, `doctor.test.ts`, `status.test.ts` related to `node:fs/promises` mocking, unrelated to this change).

## Verification

1. `git diff main -- src/commands/init.ts` shows exactly two lines removed: the discarded `readFile` call and the `readFile` symbol in the import. No other behaviour changed.
2. The `writeConfig` → `.gitignore` write → `git.commit` sequence is intact, with `await writeFile(gitignorePath, "*.tmp\n", "utf8")` now flowing directly into `await git.commit(...)`.
3. Init integration tests (`src/commands/__tests__/integration.test.ts`) continue to pass — the fresh-init, join-existing-vault, and rollback paths all observe the same on-disk state as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized the initialization process by removing a redundant file operation, improving overall performance during setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/64)

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,4 +1,4 @@
-import { mkdir, readFile, rm, writeFile } from "node:fs/promises";
+import { mkdir, rm, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import { log } from "@clack/prompts";
 import { defineCommand } from "citty";
@@ -191,7 +191,6 @@ export const initCommand = defineCommand({
 
       const gitignorePath = join(runtime.vaultDir, ".gitignore");
       await writeFile(gitignorePath, "*.tmp\n", "utf8");
-      await readFile(configPath, "utf8");
 
       const committed = await git.commit({ message: `init: ${runtime.machineName}` });
       if (committed) {


chrisleekr:

🔍 **Code review complete**, 1 files, +1/-2.

## Summary

This is a targeted dead-code removal in `src/commands/init.ts`: a discarded `await readFile(configPath, "utf8")` between the `.gitignore` write and the initial `git.commit`, plus the now-unused `readFile` import. The diff is 1 file, +1/−2. The change is correct, scoped, and risk-free. No findings — approving in spirit (a human merger still needs to give the final ack).

## What was checked

- Read `src/commands/init.ts` in full (255 lines after the change) to confirm `readFile` is not referenced anywhere else in the file. `grep -n readFile src/commands/init.ts` → 0 matches.
- Read `src/config/loader.ts` to verify the PR's contract claim: `writeConfig` calls `await mkdir(dirname(configPath), { recursive: true })` then `await writeFile(configPath, serialized, "utf8")`. If `writeConfig` resolves, the TOML file exists at `configPath`. The removed `await readFile(configPath, "utf8")` discarded its return value, so it was not validating anything — it would only have surfaced an unreadable-after-write filesystem state, which is not a realistic invariant to guard.
- Cross-referenced the surrounding flow: `writeConfig(configPath, config)` → `writeFile(gitignorePath, ...)` → `git.commit({ message: ... })`. The commit step is what actually depends on `configPath` being on disk, and it would fail loudly via `simple-git` if the file were missing. No silent failure path is introduced by removing the no-op read.
- Inspected the rollback / error-handling block (lines 229–253): the `catch` covers `writeConfig`, the `.gitignore` write, and `git.commit`/`push`, and rolls back a freshly generated key. Removing the discarded `readFile` does not change which steps the catch covers or the recovery contract documented in `ensureKeypair`.
- Verified integration coverage: `src/commands/__tests__/integration.test.ts` reads `agentsync.toml` (lines 231, 254, 714, 786) after running the init flow and asserts on its contents, so any regression where the file is not written would surface there.
- Ran `bun run typecheck` → pass (clean after `bun install`).
- Ran `bun run lint` → 8 pre-existing CSS warnings + 1 info on unrelated files; no errors, no regression on changed file.
- Ran `bun test src/commands/__tests__/integration.test.ts` → 27 pass / 0 fail (matches the PR description's claim; `src/commands/init.ts` shows 100% line coverage in the run).

## Findings

No findings. See **Reasoning** for the calls I considered.

## Reasoning

A few "this could have looked sketchy" checks I made and dismissed:

- **Was the discarded `readFile` a load-bearing fsync proxy?** No. `readFile` does not fsync; it reads through the page cache, so removing it does not change durability semantics. If the original intent was "make sure the file is flushed before `git.commit` packs it," the right tool would be `fsync`/`fdatasync` on the file descriptor, not a discarded read. The current behaviour (rely on `git.commit` to fail if the working tree is inconsistent) is what was in effect before this PR too — the read was a no-op.
- **Could the read have been masking a race with the `.gitignore` write?** No. The two writes target distinct paths (`configPath` and `gitignorePath`), and both `await` their own `writeFile`. There is no shared file handle and nothing serialized through the discarded read.
- **Does removing the read change the `catch` surface?** No. The original `readFile` could in theory have thrown (e.g., EACCES on a hostile umask) and routed into the same outer `catch` that handles the rollback. In practice `writeFile` on the same path with `"utf8"` succeeded one line earlier, so any post-write read error would be a vanishingly narrow case that is not worth a guard whose return value is thrown away. If a future change actually needs a verify-after-write, it should parse + Zod-validate via `loadConfig(configPath)`, not just read bytes.
- **Import cleanup correctness.** `readFile` is removed from the `node:fs/promises` import; `mkdir`, `rm`, and `writeFile` remain and are all still used (`mkdir` at line 128, `rm` at 73 and 238, `writeFile` at 58 and 193). No accidental orphan imports.
- **Behaviour-preserving claim.** `git diff origin/main -- src/commands/init.ts` is exactly two deletions (the call and the import symbol). No surrounding logic, ordering, or error handling was touched.

Verdict: clean dead-code removal. Safe to merge.

_cost: $0.9426 · turns: 17 · duration: 122s_

chrisleekr:

## Outstanding

Nothing blocking from this agent's side. CI is green and there are no unresolved comment threads. The only thing standing between this PR and merge is a human approval (`reviewDecision` is currently empty). Per FR-017, merging is a human action; this agent does not call `gh pr merge`.

_cost: $0.7722 · turns: 7 · duration: 48s_

PR #65, opened 2026-05-12
fix(walker): scan skill/agent bundle bodies for literal secrets

### LINKED ISSUE #57: Skill/agent bundle bodies bypass secret scanner — needs per-file scan at walker layer
## Context

Closing #47 added a central `scanForSecrets` chokepoint in `performPush` Phase 1 that scans every artifact's plaintext before encryption. The chokepoint deliberately skips artifacts whose vault path ends in `.tar.age` — skill and agent bundles. The reason: those artifacts carry **base64-encoded tar bytes**, not the underlying file contents. Running secret regexes against base64 is both:

- **unreliable** — base64 encoding scrambles credential prefixes (`sk-ant-…`, `AKIA…`, etc.), so a real secret inside a bundled markdown body would not match,
- **false-positive prone** — the base64 alphabet statistically overlaps the AWS and Google key shapes, so large bundles could abort legitimate pushes with no override.

## The remaining gap

A literal credential inside a skill's `SKILL.md`, README, or any other markdown file inside the skill directory will NOT be caught by the current chokepoint. The fix from #47 closes the gap for:

- ✅ `CLAUDE.md` and other top-level markdown.
- ✅ Copilot `*.prompt.md` files.
- ✅ Copilot `*.instructions.md` files.
- ✅ Prose-style values inside JSON `env` blocks.
- ❌ **Files inside skill / agent bundles** — still bypassed.

## Repro

```bash
echo 'token: sk-ant-api03-AAAA...' >> ~/.claude/skills/example/SKILL.md
agentsync push
# Push succeeds — the literal key is encrypted and shipped to the vault.
```

## Proposed fix

Move the secret scan **into** `collectSkillArtifacts` (`src/agents/skills-walker.ts`) and the agent-directory walk in `src/agents/copilot.ts` so each file body is scanned **before** it is added to the tar buffer. Emit warnings with the `Detected literal secret` prefix so the existing Phase-1 abort in `performPush` fires unchanged. This keeps the chokepoint contract intact while extending coverage to bundle internals.

Touch points:
- `src/agents/skills-walker.ts` — `collectSkillArtifacts` and the recursive walk inside it.
- `src/agents/copilot.ts` — the `agentsDir` walk.
- `src/agents/claude.ts` — verify no equivalent gap for plugin hook directories.

## Acceptance

- [ ] Pushing a skill that contains a `SKILL.md` with a literal Claude API key in its body aborts with a `Detected literal secret` error.
- [ ] The error message names the offending file (e.g. `~/.claude/skills/example/SKILL.md`).
- [ ] Bundles without secrets continue to push without false positives — base64 false-positive risk stays at the encoded surface, which we no longer scan.
- [ ] Tests cover both positive (secret in skill body → abort) and negative (clean skill → success).

## Out of scope

Reverting the `.tar.age` skip in `src/commands/push.ts`. Per-file scanning at the walker is the correct layer; the central scan should stay focused on whole-artifact plaintext.

chrisleekr:
## Summary

Closes #57. Adds per-file secret scanning at the walker layer so that literal credentials inside skill/agent bundle bodies (`SKILL.md`, READMEs, nested markdown, agent directory files) abort the push instead of being encrypted and shipped to the vault.

The Phase-1 chokepoint in `performPush` skips `.tar.age` artifacts on purpose — their "plaintext" is base64-encoded tar bytes, and base64 both scrambles credential prefixes (`sk-ant-…`, `AKIA…`) and false-positives against the AWS/Google key shapes. That skip closed the issue #47 gap for markdown and JSON-env bodies, but bundle internals stayed bypassed. This PR closes that remaining gap at the correct layer (the walker), without weakening the central scan's correctness boundary.

## Changes

- `src/agents/skills-walker.ts` — replace `collectNeverSyncHits` with `collectInteriorViolations`. One walk, both gates: never-sync path hits and `scanForSecrets` body hits. Each interior file is `lstat`-ed and UTF-8-read at most once; binary or unreadable files silently skip the body scan but still get the path-only never-sync check. A skill with any hit is dropped before the tar buffer is built.
- `src/agents/copilot.ts` — call `collectInteriorViolations` on each agent directory before `archiveDirectory`. The Copilot agents loop does NOT delegate to the shared skills walker, so it needs its own invocation; the helper is reused for symmetry.
- `src/commands/push.ts` — extend the walker-warning escalation loop to also escalate `Detected literal secret` (not just `never-sync inside skill: `). Update the explanatory comment block above the `.tar.age` skip to point at the new walker-layer coverage.
- `src/agents/claude.ts` — read-only audit, no code change. Top-level `commands/`, `agents/`, plugin `commands/`, plugin `agents/`, and plugin `hooks/` all emit per-file `*.md.age` / `*.json.age` artifacts whose plaintext is already scanned by the central chokepoint. Plugin-local skills route through `collectSkillArtifacts`, so they inherit the T1/T2 walker fix automatically.

## Diagram

```mermaid
sequenceDiagram
  participant W as collectSkillArtifacts<br/>(skills-walker.ts)
  participant V as collectInteriorViolations<br/>(new)
  participant T as archiveDirectory<br/>(core/tar.ts)
  participant P as performPush<br/>(commands/push.ts)
  participant S as scanForSecrets<br/>(core/sanitizer.ts)

  W->>V: walk skill interior
  V->>V: lstat + (path: shouldNeverSync)<br/>+ (body: scanForSecrets)
  alt any never-sync hit OR literal secret hit
    V-->>W: { neverSyncHits, secretWarnings }
    W-->>P: snapshot.warnings (no .tar.age artifact)
    Note over P: walker-warning loop<br/>escalates both prefixes<br/>to fatal Push aborted
  else clean
    V-->>W: empty violations
    W->>T: archiveDirectory(skill, skipSymlinks: true)
    T-->>W: tar buffer
    W-->>P: artifact (.tar.age)
    Note over P: central scan SKIPS .tar.age<br/>(base64 scramble/false-positive)
  end
```

## Files changed

- `src/agents/skills-walker.ts` · replace `collectNeverSyncHits` with `collectInteriorViolations`; add per-file `scanForSecrets`; export the new helper for cross-adapter reuse.
- `src/agents/copilot.ts` · per-file secret scan on each agent directory before `archiveDirectory`.
- `src/commands/push.ts` · escalate `Detected literal secret` walker warnings; refresh the `.tar.age` skip comment.
- `src/agents/__tests__/skills-walker.test.ts` · four new tests: positive (SKILL.md), positive (nested `notes/leak.md`), negative (AKIA-shaped prose stays included), mixed (clean + dirty).
- `src/agents/__tests__/copilot.test.ts` · two new tests covering the agents walk: agent dir with Anthropic key rejected; clean agent dir still produces one `.tar.age`.
- `src/commands/__tests__/push.test.ts` · end-to-end test: literal Claude API key in `SKILL.md` → `fatal: true`, `pushed: 0`, offending path appears in errors, no artifact written.

## Commits

- `99ef766` · fix(walker): scan skill/agent bundle bodies for literal secrets

## Tests run

- `bun run typecheck` · ✅ passes (no errors).
- `bun test src/agents/__tests__/skills-walker.test.ts src/agents/__tests__/copilot.test.ts src/commands/__tests__/push.test.ts` · ✅ **45 pass / 0 fail / 133 expect()**.
- `bun run lint` · same baseline as `main` (1 pre-existing error + 8 warnings, all in `docs/stylesheets/extra.css` — unrelated to this PR).
- `bun test` (full suite) · same baseline as `main` (239 pass / 20 fail / 16 errors). All failures are the pre-existing Bun mock-cache cross-file bleed (`SyntaxError: Missing 'default' export in module 'node:fs/promises'`) in `status.test.ts`, `daemon.test.ts`, `doctor.test.ts`, `packaging.test.ts`, etc. Verified by `git stash && bun test` on `main` returning identical counts. No new failures introduced.

## Verification

1. Walker rejects a literal `sk-ant-api03-…` pasted into `SKILL.md` with `Detected literal secret (anthropic-api-key) in <path>/SKILL.md` and drops the artifact (`skills-walker.test.ts`).
2. Walker rejects a nested `notes/leak.md` containing an `AKIA…` access key, naming the nested path (`skills-walker.test.ts`).
3. Walker still ships a clean skill whose prose contains `AKIA-history-of-aviation` and a `sha256:` digest — no false positive against the EMBEDDED_SECRET_PATTERNS (`skills-walker.test.ts`).
4. Mixed run: a clean skill is collected alongside a dirty one; only the dirty skill is dropped and warned (`skills-walker.test.ts`).
5. Copilot agents walk rejects an agent dir with a literal Anthropic key in `agent.md` and still ships clean agent dirs as a single `.tar.age` artifact (`copilot.test.ts`).
6. `performPush` end-to-end: literal key in a Copilot skill `SKILL.md` returns `fatal: true`, `pushed: 0`, includes the offending path in errors, and writes no `.tar.age` for the skill (`push.test.ts`).
7. `git grep -n "Detected literal secret" src/agents/` shows the walker now emits the contract-marker prefix, and the `.tar.age` skip in `src/commands/push.ts` is preserved (per the issue's "Out of scope").

Closes #57

## Summary by CodeRabbit

* **New Features**
  * Interior literal-secret scanning: bundles with embedded API keys or other secrets are skipped, produce clear warnings, and cause pushes to abort.

* **Bug Fixes**
  * Reduced false positives in prose scanning.
  * Duplicate warning reporting deduplicated during push aborts.

* **Tests**
  * Added unit and integration tests for interior secret detection, walker-level validation, abort-on-detection, and duplicate-warning handling.

* **Documentation**
  * Updated gating/security docs to describe combined interior scanning and push behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/65)

### DIFF
--- a/src/agents/copilot.ts
+++ b/src/agents/copilot.ts
@@ -83,13 +83,16 @@ export async function snapshotCopilot(): Promise<SnapshotResult> {
   warnings.push(...copilotSkills.warnings);
 
   // Agents directories (tar each one, similar to skills). Each agent's
-  // interior is scanned for literal credentials before the bytes head for
-  // encryption: the central scan in `commands/push.ts` skips `.tar.age`
-  // artifacts (base64 scrambles credential prefixes), so a per-file scan
-  // here is the only layer that can catch a key pasted into an agent's
-  // markdown body. Mirrors the shared walker's gate 4 contract: collect
-  // every offender, emit `Detected literal secret …` warnings (which
-  // performPush escalates to a fatal abort), and skip the artifact.
+  // interior is scanned for both never-sync path hits and literal
+  // credentials before the bytes head for encryption: the central scan
+  // in `commands/push.ts` skips `.tar.age` artifacts (base64 scrambles
+  // credential prefixes), so this per-file walk is the only layer that
+  // can catch a key pasted into an agent's markdown body or a never-sync
+  // file (auth.json, history.jsonl, sessions/**, …) nested under the
+  // agent dir. Mirrors the shared walker's gate 4 contract symmetrically:
+  // collect every offender, emit `never-sync inside skill: …` for path
+  // hits and `Detected literal secret …` for body hits — both prefixes
+  // are escalated to a fatal abort by performPush — and skip the artifact.
   try {
     const names = await readdir(AgentPaths.copilot.agentsDir);
     for (const name of names) {
@@ -98,7 +101,10 @@ export async function snapshotCopilot(): Promise<SnapshotResult> {
       if (!agentDirStat?.isDirectory()) continue;
 
       const violations = await collectInteriorViolations(agentDir);
-      if (violations.secretWarnings.length > 0) {
+      if (violations.neverSyncHits.length > 0 || violations.secretWarnings.length > 0) {
+        for (const hit of violations.neverSyncHits) {
+          warnings.push(`never-sync inside skill: ${hit}`);
+        }
         warnings.push(...violations.secretWarnings);
         continue;
       }
--- a/src/agents/copilot.ts
+++ b/src/agents/copilot.ts
@@ -5,7 +5,12 @@ import { AgentPaths } from "../config/paths";
 import { shouldNeverSync } from "../core/sanitizer";
 import { archiveDirectory, extractArchive } from "../core/tar";
 import { atomicWrite, readIfExists, type SnapshotArtifact, type SnapshotResult } from "./_utils";
-import { collectSkillArtifacts, InvalidSkillNameError, validateSkillName } from "./skills-walker";
+import {
+  collectInteriorViolations,
+  collectSkillArtifacts,
+  InvalidSkillNameError,
+  validateSkillName,
+} from "./skills-walker";
 
 /** Snapshot payload for the Copilot adapter. */
 export type CopilotSnapshotResult = SnapshotResult;
@@ -77,13 +82,27 @@ export async function snapshotCopilot(): Promise<SnapshotResult> {
   artifacts.push(...copilotSkills.artifacts);
   warnings.push(...copilotSkills.warnings);
 
-  // Agents directories (tar each one, similar to skills)
+  // Agents directories (tar each one, similar to skills). Each agent's
+  // interior is scanned for literal credentials before the bytes head for
+  // encryption: the central scan in `commands/push.ts` skips `.tar.age`
+  // artifacts (base64 scrambles credential prefixes), so a per-file scan
+  // here is the only layer that can catch a key pasted into an agent's
+  // markdown body. Mirrors the shared walker's gate 4 contract: collect
+  // every offender, emit `Detected literal secret …` warnings (which
+  // performPush escalates to a fatal abort), and skip the artifact.
   try {
     const names = await readdir(AgentPaths.copilot.agentsDir);
     for (const name of names) {
       const agentDir = join(AgentPaths.copilot.agentsDir, name);
       const agentDirStat = await stat(agentDir).catch(() => null);
       if (!agentDirStat?.isDirectory()) continue;
+
+      const violations = await collectInteriorViolations(agentDir);
+      if (violations.secretWarnings.length > 0) {
+        warnings.push(...violations.secretWarnings);
+        continue;
+      }
+
       const tarBuffer = await archiveDirectory(agentDir);
       artifacts.push({
         vaultPath: `copilot/agents/${name}.tar.age`,
--- a/src/agents/copilot.ts
+++ b/src/agents/copilot.ts
@@ -9,6 +9,7 @@ import {
   collectInteriorViolations,
   collectSkillArtifacts,
   InvalidSkillNameError,
+  NEVER_SYNC_WARNING_PREFIX,
   validateSkillName,
 } from "./skills-walker";
 
@@ -103,7 +104,10 @@ export async function snapshotCopilot(): Promise<SnapshotResult> {
       const violations = await collectInteriorViolations(agentDir);
       if (violations.neverSyncHits.length > 0 || violations.secretWarnings.length > 0) {
         for (const hit of violations.neverSyncHits) {
-          warnings.push(`never-sync inside skill: ${hit}`);
+          // Re-use the skills-walker prefix so the push gate's existing
+          // escalation rule covers agent-dir hits too — keeping the contract
+          // string in one exported constant prevents silent drift.
+          warnings.push(`${NEVER_SYNC_WARNING_PREFIX}${hit}`);
         }
         warnings.push(...violations.secretWarnings);
         continue;
--- a/src/agents/skills-walker.ts
+++ b/src/agents/skills-walker.ts
@@ -262,17 +262,27 @@ export async function collectSkillArtifacts(
  * One pass means each interior file is `lstat`-ed and read at most once,
  * and both gates share the same symlink-skipping rules.
  *
- * Symlinks (files OR sub-directories) are NOT followed and NOT inspected,
- * which is consistent with the archive step: vendored content reached via a
- * symlink is out of scope for this feature and will not be archived in gate
- * 5 either.
+ * Symlinks (files OR sub-directories) are NOT followed and NOT inspected.
+ * This matches the skills-walker caller, which then calls
+ * `archiveDirectory(skillDir, { skipSymlinks: true })`. The Copilot agents
+ * caller in `src/agents/copilot.ts` currently passes `archiveDirectory` its
+ * default (`skipSymlinks: false`) for backwards-compatibility with existing
+ * agent tarballs, so an interior symlink there is skipped by the body-scan
+ * but its target-path string is still archived. node-tar archives symlinks
+ * as `SymbolicLink` entries (target path only, not target content), so this
+ * does not exfiltrate credential bodies through a vendored helper — but
+ * callers that need full symmetry should pass `{ skipSymlinks: true }`.
  *
- * File-body reads use UTF-8. Read failures (binary files, permission errors,
- * transient I/O) silently skip the body scan for that file — the path-only
- * never-sync check still runs against the path. Binary files cannot be
- * scanned reliably with the embedded-secret regexes anyway; the goal is to
- * catch credentials pasted into text files like `SKILL.md`, READMEs, and
- * any other authored markdown or config.
+ * File-body reads use UTF-8. Permission errors and transient I/O silently
+ * skip the body scan for that file — the path-only never-sync check still
+ * runs against the path. Binary content does NOT throw under UTF-8: Node
+ * substitutes U+FFFD for invalid sequences and returns a string, so the
+ * regex still executes against the decoded bytes. This is intentional —
+ * `EMBEDDED_SECRET_PATTERNS` are anchored on real credential prefixes
+ * (`sk-ant-api03-`, `AKIA…`, `AIza…`, …) with realistic length floors, so
+ * random binary bytes virtually never match while a credential pasted into
+ * a binary blob is still caught. Do NOT add a binary-detection early-return
+ * here without a measured false-positive case; it would only weaken coverage.
  */
 export interface InteriorViolations {
   neverSyncHits: string[];
@@ -301,8 +311,12 @@ export async function collectInteriorViolations(rootDir: string): Promise<Interi
       }
 
       if (childStat.isSymbolicLink()) {
-        // Skip vendored content. The archive step filters interior symlinks
-        // out anyway, so inspecting them here would only produce noise.
+        // Skip vendored content. The skills-walker caller filters interior
+        // symlinks out of the tar archive via `skipSymlinks: true`; the
+        // Copilot agents caller currently does not, so a symlink there is
+        // archived as a `SymbolicLink` entry (path string only, not target
+        // content). See the symmetry note in `collectInteriorViolations`'s
+        // contract docstring above.
         continue;
       }
 
--- a/src/agents/skills-walker.ts
+++ b/src/agents/skills-walker.ts
@@ -45,8 +45,23 @@ export type SkillBearingAgent = "claude" | "cursor" | "codex" | "copilot";
  */
 export type SkillsWalkerResult = SnapshotResult;
 
-/** Warning prefix emitted when a never-sync rule matches inside a skill. */
-const NEVER_SYNC_WARNING_PREFIX = "never-sync inside skill: ";
+/**
+ * Warning prefix emitted when a never-sync rule matches inside a skill or
+ * agent bundle. Exported so every emitter (this file, `copilot.ts`) and the
+ * push gate (`commands/push.ts`) match the same literal — drift on this
+ * string would silently downgrade fatal aborts to soft warnings.
+ */
+export const NEVER_SYNC_WARNING_PREFIX = "never-sync inside skill: ";
+
+/**
+ * Warning prefix emitted by `scanForSecrets` when a literal credential is
+ * found inside a readable bundle interior file body. Note the trailing `(` —
+ * it disambiguates walker hits (`Detected literal secret (<name>) in …`)
+ * from redactor hits (`Detected literal secret for field <name>`), so the
+ * push gate can escalate the walker arm without re-catching the redactor
+ * warnings that the per-artifact loop already handles.
+ */
+export const WALKER_SECRET_WARNING_PREFIX = "Detected literal secret (";
 
 /** Warning prefix emitted when `archiveDirectory` fails on an otherwise-valid skill. */
 const ARCHIVE_FAILURE_WARNING_PREFIX = "skill archive failed: ";
--- a/src/agents/skills-walker.ts
+++ b/src/agents/skills-walker.ts
@@ -10,7 +10,14 @@
  *      silently so only authored skills sync.
  *   4. Interior paths matching `NEVER_SYNC_PATTERNS` emit a
  *      `never-sync inside skill: ` warning that the push pipeline escalates
- *      to a fatal abort, so secrets nested inside a skill cannot leak.
+ *      to a fatal abort, so secrets nested inside a skill cannot leak. The
+ *      same interior walk also runs `scanForSecrets` over each readable
+ *      file body and emits `Detected literal secret` warnings. The central
+ *      Phase-1 scan in `commands/push.ts` skips `.tar.age` artifacts (base64
+ *      scrambles credential prefixes and false-positives on the encoded
+ *      alphabet), so a per-file scan inside the bundle is the only layer
+ *      that can catch a literal credential pasted into `SKILL.md` or any
+ *      other interior file.
  *   5. Interior symlinks are filtered from the tar archive so vendored
  *      files cannot smuggle in through a symlinked child.
  *
@@ -18,9 +25,9 @@
  * travel back as warnings on the result so the caller decides how to surface.
  */
 
-import { lstat, readdir } from "node:fs/promises";
+import { lstat, readdir, readFile } from "node:fs/promises";
 import { join } from "node:path";
-import { shouldNeverSync } from "../core/sanitizer";
+import { scanForSecrets, shouldNeverSync } from "../core/sanitizer";
 import { archiveDirectory } from "../core/tar";
 import type { SnapshotArtifact, SnapshotResult } from "./_utils";
 
@@ -190,14 +197,20 @@ export async function collectSkillArtifacts(
     }
     if (!sentinelStat.isFile()) continue;
 
-    // Gate 4: never-sync interior walk. The walker collects every
-    // matching path in the skill — not just the first — so the user sees
-    // every offender in one push instead of fixing them one at a time.
-    const neverSyncHits = await collectNeverSyncHits(entryPath);
-    if (neverSyncHits.length > 0) {
-      for (const hit of neverSyncHits) {
+    // Gate 4: interior violation walk. A single pass collects both
+    // never-sync hits (path-pattern only) and literal secret hits
+    // (file-body scan). The walker collects every match in the skill —
+    // not just the first — so the user sees every offender in one push
+    // instead of fixing them one at a time. Literal-secret coverage at
+    // this layer closes the bundle-internals gap: the central scan in
+    // `commands/push.ts` skips `.tar.age` artifacts because base64 of a
+    // tar buffer is both scramble-prone and false-positive prone.
+    const violations = await collectInteriorViolations(entryPath);
+    if (violations.neverSyncHits.length > 0 || violations.secretWarnings.length > 0) {
+      for (const hit of violations.neverSyncHits) {
         warnings.push(`${NEVER_SYNC_WARNING_PREFIX}${hit}`);
       }
+      warnings.push(...violations.secretWarnings);
       // Skip the artifact entirely so encryption never sees the bad bytes,
       // even in the unlikely event that the push gate is later removed.
       continue;
@@ -234,15 +247,41 @@ export async function collectSkillArtifacts(
 }
 
 /**
- * Walk a real skill directory and return the absolute paths of every interior
- * file whose path matches a {@link shouldNeverSync} rule.
+ * Walk a real skill directory and surface every interior gate violation in
+ * one pass. Returned shape:
+ *
+ *   - `neverSyncHits`  — absolute paths of files whose path matches a
+ *                        {@link shouldNeverSync} rule (gate 4a, path-only).
+ *   - `secretWarnings` — `Detected literal secret …` warnings emitted by
+ *                        {@link scanForSecrets} over each readable file
+ *                        body (gate 4b, body-scan). The full warning
+ *                        string from `scanForSecrets` is preserved so the
+ *                        push pipeline's existing `Detected literal secret`
+ *                        prefix check fires unchanged.
+ *
+ * One pass means each interior file is `lstat`-ed and read at most once,
+ * and both gates share the same symlink-skipping rules.
  *
  * Symlinks (files OR sub-directories) are NOT followed and NOT inspected,
- * which is consistent with the archive step: vendored content reached via a symlink is
- * out of scope for this feature and will not be archived in gate 5 either.
+ * which is consistent with the archive step: vendored content reached via a
+ * symlink is out of scope for this feature and will not be archived in gate
+ * 5 either.
+ *
+ * File-body reads use UTF-8. Read failures (binary files, permission errors,
+ * transient I/O) silently skip the body scan for that file — the path-only
+ * never-sync check still runs against the path. Binary files cannot be
+ * scanned reliably with the embedded-secret regexes anyway; the goal is to
+ * catch credentials pasted into text files like `SKILL.md`, READMEs, and
+ * any other authored markdown or config.
  */
-async function collectNeverSyncHits(rootDir: string): Promise<string[]> {
-  const hits: string[] = [];
+export interface InteriorViolations {
+  neverSyncHits: string[];
+  secretWarnings: string[];
+}
+
+export async function collectInteriorViolations(rootDir: string): Promise<InteriorViolations> {
+  const neverSyncHits: string[] = [];
+  const secretWarnings: string[] = [];
 
   async function walk(dir: string): Promise<void> {
     let names: string[];
@@ -269,12 +308,29 @@ async function collectNeverSyncHits(rootDir: string): Promise<string[]> {
 
       if (childStat.isDirectory()) {
         await walk(childPath);
-      } else if (childStat.isFile() && shouldNeverSync(childPath)) {
-        hits.push(childPath);
+        continue;
       }
+      if (!childStat.isFile()) continue;
+
+      if (shouldNeverSync(childPath)) {
+        neverSyncHits.push(childPath);
+        // Do not also body-scan a never-sync file: it is rejected by path,
+        // and reading it would leak its contents into memory for no benefit.
+        continue;
+      }
+
+      let body: string;
+      try {
+        body = await readFile(childPath, "utf8");
+      } catch {
+        // Binary content, EACCES, or transient I/O — the path-only gate
+        // above is already best-effort for non-readable files. Skip silently.
+        continue;
+      }
+      secretWarnings.push(...scanForSecrets(body, childPath));
     }
   }
 
   await walk(rootDir);
-  return hits;
+  return { neverSyncHits, secretWarnings };
 }
--- a/src/commands/__tests__/integration.test.ts
+++ b/src/commands/__tests__/integration.test.ts
@@ -713,7 +713,9 @@ describe("integration", () => {
       const literalSecretEntries = result.errors.filter((e) => e.includes(duplicatedWarning));
       expect(literalSecretEntries.length).toBe(1);
       // The banner's count must match the distinct-issue count, not 2× it.
-      expect(result.errors[0]).toMatch(/1 security issue/);
+      // Exact substring (not a loose regex): `/1 security issue/` would also
+      // pass for `11`, `21`, … if a future regression bumped the count.
+      expect(result.errors[0]).toContain("1 security issue(s)");
     } finally {
       pushMod.__setPushAgentsForTesting([
         {
--- a/src/commands/__tests__/integration.test.ts
+++ b/src/commands/__tests__/integration.test.ts
@@ -675,6 +675,58 @@ describe("integration", () => {
     expect(result.errors.some((message) => message.includes("Detected literal secret"))).toBe(true);
   });
 
+  test("performPush does NOT double-report when a redactor warning lands on both artifact.warnings and snapshot.warnings", async () => {
+    // Real adapters (sanitizeClaudeHooks / sanitizeClaudeMcp / plugin manifest)
+    // push the `Detected literal secret for field X` warning onto BOTH the
+    // artifact (via collect()) AND the top-level snapshot.warnings (see
+    // src/agents/claude.ts:54). The Phase-1 artifact loop and the walker-
+    // warning loop must not both fire on the same warning, or one secret would
+    // produce two entries in the abort banner. Walker hits use the
+    // `Detected literal secret (<name>) in …` shape; redactor hits use
+    // `Detected literal secret for field <name>` — the snapshot-level prefix
+    // matches the walker shape only.
+    const duplicatedWarning = "Detected literal secret for field apiKey";
+    const dualSnapshotAgent = [
+      {
+        name: "claude" as const,
+        snapshot: async () => ({
+          artifacts: [
+            {
+              vaultPath: "claude/settings.age",
+              sourcePath: "/fake/.claude/settings.json",
+              plaintext: '{"apiKey":"[REDACTED]"}',
+              warnings: [duplicatedWarning],
+            },
+          ],
+          warnings: [duplicatedWarning],
+        }),
+        apply: async () => {},
+      },
+    ];
+
+    pushMod.__setPushAgentsForTesting(dualSnapshotAgent);
+    try {
+      const result = await pushMod.performPush({ agent: "claude" });
+
+      expect(result.fatal).toBe(true);
+      expect(result.pushed).toBe(0);
+      const literalSecretEntries = result.errors.filter((e) => e.includes(duplicatedWarning));
+      expect(literalSecretEntries.length).toBe(1);
+      // The banner's count must match the distinct-issue count, not 2× it.
+      expect(result.errors[0]).toMatch(/1 security issue/);
+    } finally {
+      pushMod.__setPushAgentsForTesting([
+        {
+          name: "claude" as const,
+          snapshot: async () => ({ artifacts: [...fakeArtifacts], warnings: [] }),
+          apply: async (vaultDir: string) => {
+            fakeApplyCalls.push(vaultDir);
+          },
+        },
+      ]);
+    }
+  });
+
   test("status command runs without throwing", async () => {
     const statusMod = await import("../../commands/status");
     await statusMod.statusCommand.run?.({
--- a/src/commands/push.ts
+++ b/src/commands/push.ts
@@ -4,6 +4,7 @@ import { log } from "@clack/prompts";
 import { defineCommand } from "citty";
 import { applyClaudeVault, type ClaudeSyncOptions, snapshotClaude } from "../agents/claude";
 import { type AgentDefinition, type AgentName, Agents } from "../agents/registry";
+import { NEVER_SYNC_WARNING_PREFIX, WALKER_SECRET_WARNING_PREFIX } from "../agents/skills-walker";
 import { encryptString } from "../core/encryptor";
 import { GitClient } from "../core/git";
 import { scanForSecrets, shouldNeverSync } from "../core/sanitizer";
@@ -148,7 +149,7 @@ export async function performPush(
     // snapshot.warnings; the per-artifact loop above already catches them, so
     // a broad `Detected literal secret` match here would double-report.
     for (const w of snapshot.warnings) {
-      if (w.startsWith("never-sync inside skill: ") || w.startsWith("Detected literal secret (")) {
+      if (w.startsWith(NEVER_SYNC_WARNING_PREFIX) || w.startsWith(WALKER_SECRET_WARNING_PREFIX)) {
         secretErrors.push(`[${agent.name}] ${w}`);
       }
     }
--- a/src/commands/push.ts
+++ b/src/commands/push.ts
@@ -115,8 +115,14 @@ export async function performPush(
       // statistically overlaps with `AKIA…` and `AIza…` credentials, so
       // scanning the encoded form would false-positive without reliably
       // catching credentials *inside* the bundle (encoding scrambles
-      // prefixes). Per-file scanning at the walker layer is tracked
-      // separately; for now skip the encoded surface entirely.
+      // prefixes). The encoded surface is intentionally skipped here.
+      // Bundle internals are covered separately at the walker layer:
+      // `collectInteriorViolations` in `src/agents/skills-walker.ts` scans
+      // each readable interior file body with `scanForSecrets` before the
+      // tar buffer is built, and the Copilot agents walk in
+      // `src/agents/copilot.ts` invokes the same helper. Both surface
+      // `Detected literal secret …` warnings on the snapshot, which the
+      // walker-warning loop below escalates to a fatal abort.
       if (artifact.vaultPath.endsWith(".tar.age")) {
         continue;
       }
@@ -124,13 +130,18 @@ export async function performPush(
         secretErrors.push(`[${agent.name}] ${w}`);
       }
     }
-    // Walker-level warnings (e.g. "never-sync inside skill: <path>") are
-    // emitted on the top-level snapshot.warnings array by the shared skills
-    // walker (src/agents/skills-walker.ts), not on individual artifacts. They
-    // must also escalate to a fatal abort so a never-sync file inside a skill
-    // directory never reaches encryption.
+    // Walker-level warnings are emitted on the top-level snapshot.warnings
+    // array (not per-artifact, since the offending bundle is dropped before
+    // any artifact is built). Two prefixes escalate to a fatal abort:
+    //   - `never-sync inside skill: <path>` — a path-pattern hit inside a
+    //     skill, so the bundle would have contained a hard never-sync file.
+    //   - `Detected literal secret …` — a credential found inside an
+    //     interior file body during the walker's per-file scan. The central
+    //     scan a few lines above skips `.tar.age` artifacts, so this is the
+    //     only layer that catches a key pasted into `SKILL.md`, READMEs, or
+    //     any other file inside a skill/agent bundle.
     for (const w of snapshot.warnings) {
-      if (w.startsWith("never-sync inside skill: ")) {
+      if (w.startsWith("never-sync inside skill: ") || w.startsWith("Detected literal secret")) {
         secretErrors.push(`[${agent.name}] ${w}`);
       }
     }
--- a/src/commands/push.ts
+++ b/src/commands/push.ts
@@ -135,13 +135,20 @@ export async function performPush(
     // any artifact is built). Two prefixes escalate to a fatal abort:
     //   - `never-sync inside skill: <path>` — a path-pattern hit inside a
     //     skill, so the bundle would have contained a hard never-sync file.
-    //   - `Detected literal secret …` — a credential found inside an
-    //     interior file body during the walker's per-file scan. The central
-    //     scan a few lines above skips `.tar.age` artifacts, so this is the
-    //     only layer that catches a key pasted into `SKILL.md`, READMEs, or
-    //     any other file inside a skill/agent bundle.
+    //   - `Detected literal secret (<name>) in <path>` — a credential found
+    //     inside an interior file body during the walker's per-file scan.
+    //     The central scan a few lines above skips `.tar.age` artifacts, so
+    //     this is the only layer that catches a key pasted into `SKILL.md`,
+    //     READMEs, or any other file inside a skill/agent bundle.
+    //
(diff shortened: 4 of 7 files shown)


chrisleekr:

🔍 **Code review complete**, 6 files, +322/-28.

## Summary

PR #65 closes the bundle-internals literal-secret gap by adding a per-file `scanForSecrets` walk inside `collectInteriorViolations` (formerly `collectNeverSyncHits`) and wiring the `Detected literal secret` prefix into the existing walker-warning escalation in `performPush`. The core walker change, the push-side prefix extension, and the new tests are well-scoped and correct. **One major finding**: the new Copilot agents-loop call site consumes only half of what the walker returns — `neverSyncHits` are computed and then discarded, leaving the pre-existing never-sync gap for `~/.copilot/agents/<name>/auth.json` (and the other `NEVER_SYNC_PATTERNS` entries) open even though the walker now reports it for free. Otherwise the change is sound and ships net-positive security coverage. Verdict: request fix to the agents loop before merge; the rest is approved on substance.

## What was checked

- `git diff origin/main...HEAD` — full diff for all 6 files.
- Read in full:
  - `src/agents/skills-walker.ts` (the new `collectInteriorViolations` helper, the renamed gate 4 in `collectSkillArtifacts`, the contract docstring on the returned shape)
  - `src/agents/copilot.ts` (the agents-loop integration and the symmetric `applyCopilot*` callers)
  - `src/commands/push.ts` (the `Detected literal secret` escalation and the refreshed `.tar.age` skip comment)
  - `src/agents/__tests__/skills-walker.test.ts` (four new tests covering positive SKILL.md hit, nested file hit, false-positive negative, mixed clean+dirty)
  - `src/agents/__tests__/copilot.test.ts` (two new tests on the agents walk)
  - `src/commands/__tests__/push.test.ts` (end-to-end abort test)
- Cross-referenced `collectInteriorViolations` / `collectNeverSyncHits` callers (3 files), `archiveDirectory` callers (15+ usages across agents and tests), `scanForSecrets` / `Detected literal secret` prefix producers and consumers, and `NEVER_SYNC_PATTERNS` membership in `src/core/sanitizer.ts`.
- Compared pre-PR `src/agents/copilot.ts` to post-PR to confirm whether the never-sync gap pre-existed (it did; the agents loop never had any interior gate) or was introduced (it wasn't — but the new walk now reports the hits and the call site explicitly throws them away).
- Validation:
  - `bun install` then `bun run typecheck` — clean, no errors.
  - `bun test src/agents/__tests__/skills-walker.test.ts src/agents/__tests__/copilot.test.ts src/commands/__tests__/push.test.ts` — **45 pass / 0 fail / 133 expect()**, matches the PR description.
  - Empirical probe: a Copilot agent dir seeded with `agent.md` + `auth.json` produced `copilot/agents/<name>.tar.age` with zero warnings, confirming the gap.

## Findings

### [major] `src/agents/copilot.ts:100-104` — Copilot agents loop discards `violations.neverSyncHits`

`collectInteriorViolations` returns both `neverSyncHits` and `secretWarnings`. The agents loop only checks `secretWarnings`, so files matching `NEVER_SYNC_PATTERNS` (`auth.json`, `.credentials.json`, `history.jsonl`, `sessions/**`, `*.local.md`, `.claude/settings.local.json`, `agentsync.toml`, `*.age`) inside `~/.copilot/agents/<name>/` are detected by the walker, then passed through `archiveDirectory(agentDir)` (no never-sync filter; no `skipSymlinks`) and shipped to the vault. Pre-PR there was no interior gate at all, so this is not a regression of behavior — but the PR's stated symmetry ("Mirrors the shared walker's gate 4 contract"), the comment block above, and the fact that the walker now produces the answer **for free**, make this the right PR to close it. The skills walker handles both arms at `src/agents/skills-walker.ts:209-217`.

**Recommended fix:** Mirror the skills walker — `if (violations.neverSyncHits.length > 0 || violations.secretWarnings.length > 0) { … push neverSyncHits with the "never-sync inside skill: " prefix; push secretWarnings as-is; continue; }`. The `commands/push.ts:144` gate already escalates that prefix to a fatal abort. Add a regression test (`auth.json` in a Copilot agent dir → no `.tar.age` written, push aborted). One inline comment posted at `src/agents/copilot.ts:104`.

## Reasoning

Several aspects of the diff looked sketchy at first read but check out on inspection:

- **`scanForSecrets` is invoked on every readable interior file** — could it false-positive on hash digests, lockfile blobs, or AKIA-shaped prose? No. `EMBEDDED_SECRET_PATTERNS` (`src/core/sanitizer.ts:74-83`) are anchored on credential-prefix tokens with realistic length floors (`AKIA[0-9A-Z]{16}`, `AIza[0-9A-Za-z_-]{35}`, `sk-ant-api03-[A-Za-z0-9_-]{40,}`, etc.); `sha256:abcd...` (48 hex) and `AKIA-history-of-aviation` cannot match. The PR test `does NOT false-positive on a clean skill with AKIA-shaped prose` exercises both cases.
- **Binary files getting silently mis-scanned** — `readFile(path, "utf8")` does not throw on binary input; it returns a U+FFFD-laden string and the regex still runs. The comment at `skills-walker.ts:322-328` implies the catch handles binary content, which is slightly off, but the regex specificity means it is harmless in practice (a random byte sequence will not produce `sk-ant-api03-` + 40 valid base64url chars). Comment nit at most; not flagged.
- **Path-traversal / vault basename safety** — unchanged by this PR. `validateSkillName` (rejecting `.`, `..`, separators, control chars, leading dot) is already applied symmetrically to skills and agents on the pull side via `applyCopilotSkill` and `applyCopilotAgent`, with adversarial-filename tests already in `copilot.test.ts`.
- **Prefix collision between walker `scanForSecrets` ("Detected literal secret (<name>) in <path>") and JSON-redactor `redactSecretLiterals` ("Detected literal secret for field <name>")** — both start with `Detected literal secret`, and the push gate's `.startsWith("Detected literal secret")` (no trailing space) intentionally catches both. Correct.
- **Walker still ships the clean skill in mixed runs** — confirmed by tests `Row 12` and `collects clean skills even when another skill has a literal-secret hit`. The R3 additive contract is preserved.
- **Phase-1 chokepoint deliberately skips `.tar.age`** — the rationale (base64 scrambles credential prefixes, alphabet overlaps `AKIA…`/`AIza…`) is correct, and the refreshed comment in `push.ts:108-128` now points readers to the walker-layer coverage. The two-layer model is internally consistent.

The single deliberate gap I am flagging is the agents-loop omission. Closing it costs ~2 lines + 1 regression test and brings the agents walk into actual parity with the skills walk that the comment already claims.

_cost: $3.2000 · turns: 36 · duration: 348s_

src/agents/copilot.ts
```diff
@@ -77,13 +82,27 @@ export async function snapshotCopilot(): Promise<SnapshotResult> {
...
     for (const name of names) {
       const agentDir = join(AgentPaths.copilot.agentsDir, name);
       const agentDirStat = await stat(agentDir).catch(() => null);
       if (!agentDirStat?.isDirectory()) continue;
+
+      const violations = await collectInteriorViolations(agentDir);
+      if (violations.secretWarnings.length > 0) {
+        warnings.push(...violations.secretWarnings);
+        continue;
+      }
```
chrisleekr, line 114: _⚠️ Potential issue_ | _🟠 Major_

**Copilot agents walk silently drops `violations.neverSyncHits` even though the helper already computed them.**

`collectInteriorViolations` returns both `neverSyncHits` and `secretWarnings`, but `src/agents/copilot.ts:100-104` only inspects `secretWarnings`. An agent dir containing a `NEVER_SYNC_PATTERNS` file (`auth.json`, `.credentials.json`, `history.jsonl`, `sessions/**`, `*.local.md`, etc.) is detected by the walker, then archived with `archiveDirectory(agentDir)` (no skip-symlinks, no never-sync filter) and shipped — exactly the credential class that `auth.json` is on the list to keep out of the vault. Verified empirically with a probe: an agent dir holding `agent.md` + `auth.json` produced `copilot/agents/<name>.tar.age` with zero warnings. The skills walker handles both gates symmetrically at `src/agents/skills-walker.ts:209-217`; the comment above this loop also claims to "Mirror the shared walker's gate 4 contract", but the mirror is missing the never-sync arm.

**Suggested fix:** Mirror the skills walker check — `if (violations.neverSyncHits.length > 0 || violations.secretWarnings.length > 0)`, push each `neverSyncHits` entry with the `never-sync inside skill: ` prefix (already escalated to a fatal abort by `commands/push.ts:144`) alongside `secretWarnings`, then `continue`. Add a regression test with `auth.json` inside a Copilot agent dir to lock it in.

chrisleekr:

🔎 **Resolve incomplete**, 1 checks still in flight after the agent finished.

## Outstanding

- **Lint & Typecheck CI job hung.** The job has been `in_progress` since 10:52:10 UTC with no step output. Other checks on the same head SHA all pass, and local `bun run typecheck` plus the lefthook pre-commit typecheck both succeeded. This appears to be a GitHub-Actions runner / queue stall unrelated to the diff. Recommended maintainer action: re-run the workflow (`gh run rerun 25729747044 --failed` once it eventually times out, or cancel + re-trigger via an empty `--allow-empty` commit). Do not merge until this check is green — `mergeStateStatus` is `BLOCKED` until then.
- No other blockers. `reviewDecision` is empty (no formal approval yet); both review threads have been addressed.

_cost: $3.0432 · turns: 46 · duration: 2715s_

src/agents/copilot.ts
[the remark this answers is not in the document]
chrisleekr: _✅ Addressed_, commit `eb5e2f1`

**Copilot agents loop now checks both arms of `collectInteriorViolations`.**

Mirrored the skills walker at `src/agents/copilot.ts:104`: when either `neverSyncHits` or `secretWarnings` is non-empty, drop the artifact and emit each `neverSyncHits` entry with the existing `never-sync inside skill: ` prefix (already escalated to a fatal abort by `commands/push.ts:144`) alongside the secret warnings. Added a regression test seeding `auth.json` inside an agent dir, asserting the `.tar.age` artifact is not written and the never-sync warning surfaces.

chrisleekr: _✅ Addressed_, commit `eb5e2f1`

**Agents loop now consumes both arms of the helper; never-sync hits abort the push.**

Updated `src/agents/copilot.ts:100-110` to mirror the skills walker at `src/agents/skills-walker.ts:209-217`: both `neverSyncHits` and `secretWarnings` are now inspected, never-sync hits are emitted with the `never-sync inside skill: ` prefix that `commands/push.ts:144` already escalates to a fatal abort, and the artifact is dropped before `archiveDirectory` runs. Added a regression test in `src/agents/__tests__/copilot.test.ts` ("agent dir containing a never-sync file (auth.json) is rejected") that seeds `auth.json` inside an agent dir and asserts no `.tar.age` is produced. The pre-existing leak path you flagged with the empirical probe is now closed.

chrisleekr:

🔍 **Code review complete**, 6 files, +354/-28.

## Summary

PR #65 closes the bundle-internals literal-secret gap by adding a per-file `scanForSecrets` walk inside the new `collectInteriorViolations` helper (formerly `collectNeverSyncHits`) and wiring the `Detected literal secret` prefix into the existing walker-warning escalation in `performPush`. The core walker change, the Copilot agents-loop integration (already strengthened in commit `eb5e2f1` to consume both `neverSyncHits` and `secretWarnings`), and the new tests are well-scoped and correct. **One minor regression** worth fixing before merge: the broadened snapshot-level prefix match in `src/commands/push.ts:144` double-reports the `Detected literal secret for field X` warnings that `redactSecretLiterals`-based sanitizers already emit on `artifact.warnings`, so a single literal secret in `~/.claude/settings.json` now produces two duplicate entries in `secretErrors`. Plus two docstring nits in `src/agents/skills-walker.ts` (the "consistent with gate 5" symmetry claim no longer holds for the Copilot agents caller, and the "binary files silently skip the body scan" claim is materially wrong — `readFile(utf8)` does not throw on binary). Verdict: ship with the duplicate-report fix; the docstring issues can ride along or as a follow-up.

## What was checked

- `git diff origin/main...HEAD` — full diff for all 6 files (`+354/-28`).
- Read in full:
  - `src/agents/skills-walker.ts` — new `InteriorViolations` interface, exported `collectInteriorViolations` helper, the renamed gate 4 invocation in `collectSkillArtifacts`.
  - `src/agents/copilot.ts` — agents-loop integration of `collectInteriorViolations` consuming BOTH `neverSyncHits` and `secretWarnings`.
  - `src/commands/push.ts` — broadened snapshot-warning escalation to include `Detected literal secret`, refreshed comment block above the `.tar.age` skip.
  - `src/agents/__tests__/skills-walker.test.ts` — four new tests (positive SKILL.md hit, nested file hit, false-positive negative on AKIA-shaped prose + sha256 digests, mixed clean+dirty).
  - `src/agents/__tests__/copilot.test.ts` — three new tests on the agents walk (literal Anthropic key, `auth.json` never-sync hit, clean-dir baseline).
  - `src/commands/__tests__/push.test.ts` — end-to-end abort test via `performPush({ agent: "copilot" })`.
- Cross-referenced `collectInteriorViolations` / `collectNeverSyncHits` callers (3 sites), `archiveDirectory` callers and the `skipSymlinks` default (`src/core/tar.ts:8-18`), `scanForSecrets` and `Detected literal secret` prefix producers (`src/core/sanitizer.ts:117-125` and `:128-138`) and consumers (`src/commands/push.ts:104,144`).
- Compared `claude.ts`/`cursor.ts`/`codex.ts` to confirm the Copilot agents loop is the only `archiveDirectory` consumer outside the skills walker — the others use per-file `.md.age` artifacts that already flow through the central Phase-1 scan.
- Validation:
  - `bun install` — 38 packages installed.
  - `bun test src/agents/__tests__/skills-walker.test.ts src/agents/__tests__/copilot.test.ts src/commands/__tests__/push.test.ts` — **46 pass / 0 fail / 136 expect()** (PR description said 45/133; the extra test + expects come from the `auth.json` regression test added in `eb5e2f1`).
  - `bun run lint` — baseline matches `main` (1 info + 8 warnings, all in `docs/stylesheets/extra.css`).
  - Empirical probe in-repo: `sanitizeClaudeHooks({hooks:{onSomething:{apiKey:"ghp_…"}}})` returns `["Detected literal secret for field apiKey"]`, confirming the duplicate-escalation finding.
  - Empirical probe: `readFile("/tmp/binary-with-key.bin", "utf8")` on a buffer holding `0x00 0xff 0xfe 0x80 0x81 0x82 "sk-ant-api03-…" 0x80 0x00 0xff` does NOT throw — returns a 70-char string — and the regex matches, confirming the binary-file docstring finding.

## Findings

### [minor] `src/commands/push.ts:144` — duplicate escalation on `Detected literal secret for field X`

The broadened snapshot-level prefix match now catches `Detected literal secret for field <name>` warnings emitted by `redactSecretLiterals` inside `sanitizeClaudeHooks`/`sanitizeClaudeMcp`/`sanitizeClaudePluginManifest`/`sanitizeClaudePluginMcp`. Those warnings are already on `artifact.warnings` (via `collect()` in `src/agents/_utils.ts:65-68`) AND `snapshot.warnings` (e.g. `src/agents/claude.ts:54,61,133,152,192,277`). Pre-PR the artifact-level loop at `push.ts:104` caught them once and the snapshot-level loop only matched `never-sync inside skill:`. Post-PR they fire in both loops, so one literal secret in `~/.claude/settings.json` produces two identical entries in `secretErrors` and the user-visible banner reads `Push aborted: 2 security issue(s) detected.` for one secret. Functionally correct (push still aborts), but a real UX regression with no existing test that catches it (`integration.test.ts:661` puts the warning only on the artifact, not on the snapshot).

**Recommended fix:** narrow the snapshot-level prefix to the walker-specific shape `Detected literal secret (` (open paren — only `scanForSecrets` emits that form; `redactSecretLiterals` emits `Detected literal secret for field …`), or de-duplicate `secretErrors` before formatting. Inline comment posted at `src/commands/push.ts:144`.

### [minor] `src/agents/skills-walker.ts:265-268` — docstring "consistent with the archive step" no longer holds for the Copilot agents caller

The helper's docstring promises symlink-skip symmetry with "gate 5" (`archiveDirectory(…, { skipSymlinks: true })`). That's true for the skills walker caller at line 222 but false for the new Copilot agents caller at `src/agents/copilot.ts:112`, which calls `archiveDirectory(agentDir)` without `skipSymlinks: true`. Per `src/core/tar.ts:13` the default ("preserves the existing tar-everything behavior used by Copilot agent-tarballs") leaves symlinks in the archive. node-tar archives them as `SymbolicLink` entries — only the path string, not the target's content — so this is not a credential-content leak, but a vendored helper symlinked into an agent dir bypasses the credential-body scan AND ships its target path string to the vault, contradicting the documented symmetry.

**Recommended fix:** either pass `{ skipSymlinks: true }` to `archiveDirectory` in `copilot.ts:112` to make the helper's promise true for both callers, or scope the comment to the skills walker caller and call out the asymmetry. Inline comment posted at `src/agents/skills-walker.ts:268`.

### [nit] `src/agents/skills-walker.ts:270-275` — docstring claim about binary files is materially wrong

The docstring says "Read failures (binary files, permission errors, transient I/O) silently skip the body scan". `readFile(path, "utf8")` does NOT throw on binary content — Node replaces invalid UTF-8 with U+FFFD and returns a string; the regex still runs. Empirically verified in-repo: a buffer containing `0x00 0xff 0xfe 0x80 0x81 0x82 "sk-ant-api03-…" 0x80 0x00 0xff` reads fine as a 70-char string and the embedded-secret regex matches. The current behavior is actually correct (a credential pasted into a binary blob is still caught thanks to prefix specificity), but the docstring's framing could mislead a future maintainer into "fixing" the non-bug by adding a binary-skip early-return that would weaken coverage.

**Recommended fix:** correct the comment to reflect reality — binary files ARE scanned and the high prefix-specificity of `EMBEDDED_SECRET_PATTERNS` keeps false positives near zero. Inline comment posted at `src/agents/skills-walker.ts:275`.

## Reasoning

Several aspects of the diff looked sketchy at first read but check out on inspection:

- **`scanForSecrets` false-positives on hash digests / lockfile blobs / AKIA-shaped prose** — `EMBEDDED_SECRET_PATTERNS` are anchored on real credential-prefix tokens with realistic length floors (`AKIA[0-9A-Z]{16}`, `AIza[0-9A-Za-z_-]{35}`, `sk-ant-api03-[A-Za-z0-9_-]{40,}`, etc.); a `sha256:` digest of 48 hex chars and `AKIA-history-of-aviation` (the trailing chars include `-`, breaking `[0-9A-Z]`) cannot match. The PR test "does NOT false-positive on a clean skill with AKIA-shaped prose" exercises both.
- **Phase-1 chokepoint deliberately skips `.tar.age`** — the rationale (base64 scrambles credential prefixes and the alphabet overlaps `AKIA…`/`AIza…`) is correct, and the refreshed comment in `push.ts:108-128` correctly points readers at the walker-layer coverage. The two-layer model is internally consistent.
- **Walker still ships the clean skill in mixed runs** — confirmed by the "Row 12" test and the new "collects clean skills even when another skill has a literal-secret hit" test. The additive contract is preserved.
- **Path-traversal / vault basename safety** — unchanged by this PR. `validateSkillName` (rejecting `.`, `..`, separators, control chars, leading dot) is already applied symmetrically to skills and agents on the pull side via `applyCopilotSkill`/`applyCopilotAgent`, with adversarial-filename tests in `copilot.test.ts:438-504`.
- **Copilot agents loop discarding `neverSyncHits`** — the prior review thread flagged this on commit `99ef766`; the maintainer addressed it in `eb5e2f1`, which is the head of this branch. Current code at `copilot.ts:104-110` correctly inspects both arms, emits `never-sync inside skill: …` (re-using the existing escalation prefix), and continues. The new regression test "agent dir containing a never-sync file (auth.json) is rejected" locks it in.
- **Claude / Cursor / Codex** — `claude.ts:88-108` emits per-file `claude/agents/<name>.md.age` (NOT `.tar.age`), so the central Phase-1 scan at `push.ts:129` already catches credentials in their bodies. `cursor.ts` and `codex.ts` have no `archiveDirectory` callers. Their omission from this PR is correct.
- **Silent try/catch around the Copilot agents loop** (`copilot.ts:96-122`) — pre-existing pattern (compare `git show origin/main:src/agents/copilot.ts`); not introduced by this PR. If `collectInteriorViolations` or `archiveDirectory` ever threw mid-loop the remaining agents would be silently skipped, but `collectInteriorViolations` is documented as never-throwing and the pre-PR `archiveDirectory` path had the same fragility.
- **Naming nit on the `never-sync inside skill:` prefix used for an agent dir** (`copilot.ts:106`) — semantically inaccurate but pragmatically right: the existing `push.ts:144` gate matches that exact prefix and a rename would require a coordinated change to both call sites with no behavioral benefit. The prior review thread accepted this framing.

The single deliberate UX regression I am flagging is the duplicate-escalation on push.ts:144 — straightforward to fix with a more specific prefix, and worth handling before merge so the abort banner counts and lists are not artificially doubled.

_cost: $5.5634 · turns: 64 · duration: 563s_

src/commands/push.ts
```diff
@@ -115,22 +115,33 @@ export async function performPush(
...
+    //   - `never-sync inside skill: <path>` — a path-pattern hit inside a
+    //     skill, so the bundle would have contained a hard never-sync file.
+    //   - `Detected literal secret …` — a credential found inside an
+    //     interior file body during the walker's per-file scan. The central
+    //     scan a few lines above skips `.tar.age` artifacts, so this is the
+    //     only layer that catches a key pasted into `SKILL.md`, READMEs, or
+    //     any other file inside a skill/agent bundle.
     for (const w of snapshot.warnings) {
-      if (w.startsWith("never-sync inside skill: ")) {
+      if (w.startsWith("never-sync inside skill: ") || w.startsWith("Detected literal secret")) {
```
chrisleekr: _💭 Suggestion_      | _🟡 Minor_

**Broadened snapshot-level prefix match double-reports `Detected literal secret for field X` warnings already caught at the artifact level.**

`sanitizeClaudeHooks`, `sanitizeClaudeMcp`, `sanitizeClaudePluginManifest`, etc. emit `Detected literal secret for field <name>` warnings via `redactSecretLiterals` and push them onto BOTH `artifact.warnings` (through `collect()` in `src/agents/_utils.ts:65-68`) AND `snapshot.warnings` (e.g. `src/agents/claude.ts:54,61,133,152,192,277`). Pre-PR, the artifact-level loop at `src/commands/push.ts:104` caught the warning once and the snapshot-level loop only matched `never-sync inside skill:`. After this PR the same warning is also matched on line 144, so a single literal secret in `~/.claude/settings.json` produces two identical entries in `secretErrors`. User-visible result: `Push aborted: 2 security issue(s) detected.` for one secret, with the same offending field listed twice. Functionally correct (push still aborts), but a real UX regression and no existing test exercises it (`integration.test.ts:661` puts the warning only on the artifact). Empirically confirmed: `sanitizeClaudeHooks({hooks:{onSomething:{apiKey:"ghp_..."}}})` returns `["Detected literal secret for field apiKey"]`, and the adapter dual-pushes it.

**Suggested fix:** Narrow the snapshot-level prefix to the walker's exact shape — `Detected literal secret (` (note the open paren that only `scanForSecrets` emits, never `redactSecretLiterals`) — so the walker path stays escalated without re-catching warnings the artifact loop already handles. Alternatively, de-duplicate `secretErrors` before formatting.

src/agents/skills-walker.ts
```diff
@@ -234,15 +247,41 @@ export async function collectSkillArtifacts(
...
+ *
+ * One pass means each interior file is `lstat`-ed and read at most once,
+ * and both gates share the same symlink-skipping rules.
  *
  * Symlinks (files OR sub-directories) are NOT followed and NOT inspected,
- * which is consistent with the archive step: vendored content reached via a symlink is
- * out of scope for this feature and will not be archived in gate 5 either.
+ * which is consistent with the archive step: vendored content reached via a
+ * symlink is out of scope for this feature and will not be archived in gate
+ * 5 either.
```
chrisleekr: _💭 Suggestion_      | _🟡 Minor_

**Docstring promise "consistent with the archive step / will not be archived in gate 5 either" no longer holds for the Copilot agents caller.**

The helper is now exported and consumed by `src/agents/copilot.ts:103`, which then calls `archiveDirectory(agentDir)` at line 112 — without `{ skipSymlinks: true }`. The default `skipSymlinks: false` (documented at `src/core/tar.ts:13` as "preserves the existing tar-everything behavior used by Copilot agent-tarballs") means symlinks inside an agent dir ARE archived. So for the Copilot agents walk, the walker silently skips a symlinked file at line 303-307 AND the archive step preserves it — the opposite of what the docstring promises. node-tar archives symlinks as `SymbolicLink` entries (target path, not content), so this isn't a credential-content leak in practice; but a vendored helper symlinked into an agent dir bypasses the credential-body scan AND ships its target path string to the vault, contradicting the documented symmetry that the skills-walker caller relies on.

**Suggested fix:** Either add `{ skipSymlinks: true }` to `archiveDirectory(agentDir, …)` in `src/agents/copilot.ts:112` to make the helper's symmetry promise true for both callers, or rephrase the comment here (lines 265-268) to scope the "consistent with gate 5" claim to the skills-walker caller only and call out the Copilot-agents asymmetry explicitly.

src/agents/skills-walker.ts
```diff
@@ -234,15 +247,41 @@ export async function collectSkillArtifacts(
...
+ *
+ * File-body reads use UTF-8. Read failures (binary files, permission errors,
+ * transient I/O) silently skip the body scan for that file — the path-only
+ * never-sync check still runs against the path. Binary files cannot be
+ * scanned reliably with the embedded-secret regexes anyway; the goal is to
+ * catch credentials pasted into text files like `SKILL.md`, READMEs, and
+ * any other authored markdown or config.
```
chrisleekr: _💭 Suggestion_      | _🔵 Nit_

**Docstring claim that binary files "silently skip the body scan" is materially wrong; the regex still runs on them.**

`readFile(childPath, "utf8")` does NOT throw on binary content — Node returns a string where invalid UTF-8 byte sequences are replaced with U+FFFD, and the `scanForSecrets` regex still executes against that string. Empirically verified with a buffer of `[0x00 0xff 0xfe 0x80 0x81 0x82 "sk-ant-api03-..." 0x80 0x00 0xff]`: `readFile` returns a 70-char string and `/sk-ant-api03-[A-Za-z0-9_-]{40,}/.test(body)` returns `true`. So binary files ARE scanned (which is actually a security WIN — a credential embedded in a binary blob is still caught thanks to the prefix specificity in `EMBEDDED_SECRET_PATTERNS`). The docstring's framing risks misleading a future maintainer into "fixing" the non-bug by adding an explicit binary-detection early-return, which would weaken coverage.

**Suggested fix:** Update the comment block at lines 270-275 to reflect reality — something like: "Read failures (permission errors, transient I/O) silently skip the body scan. Binary content reads fine under UTF-8 with replacement characters and IS scanned; the embedded-secret regexes are specific enough that random binary bytes do not false-positive in practice."

chrisleekr:

## Outstanding

Nothing blocks merge. All five required checks are green. `CodeRabbit` is still running its (optional) review pass — `mergeStateStatus: UNSTABLE` will resolve to `CLEAN` once it finishes. No formal approval (`reviewDecision: null`) — that's a human action.

_cost: $4.6760 · turns: 72 · duration: 472s_

src/commands/push.ts
chrisleekr: _✅ Addressed_, commit `944eff3`

**Narrowed snapshot-level walker-secret prefix so redactor warnings are not double-reported.**

You were right — `sanitizeClaudeHooks`/`Mcp`/`PluginManifest`/`PluginMcp` push `Detected literal secret for field <name>` onto both `artifact.warnings` (via `collect()`) and `snapshot.warnings`, so the broadened match at `src/commands/push.ts:144` was catching them twice (once in the per-artifact loop at `:104`, once in the snapshot-warning loop). Tightened the snapshot loop to match `Detected literal secret (` (open paren — only `scanForSecrets` emits that shape, never `redactSecretLiterals`), preserving the walker-layer escalation without re-catching the redactor path. Added a regression test in `src/commands/__tests__/integration.test.ts` that injects a fake agent whose snapshot emits the same redactor warning on BOTH `artifact.warnings` and `snapshot.warnings`, then asserts the abort banner says `1 security issue` (not `2`) and `secretErrors` contains exactly one matching entry.

src/agents/skills-walker.ts
chrisleekr: _✅ Addressed_, commit `944eff3`

**Rewrote the symlink-symmetry docstring to call out the Copilot agents asymmetry explicitly.**

You're correct — the Copilot agents caller in `src/agents/copilot.ts:112` passes `archiveDirectory(agentDir)` with the default `skipSymlinks: false`, so the helper's "consistent with gate 5" promise only holds for the skills-walker caller. Updated the contract docstring in `src/agents/skills-walker.ts` to scope the symmetry claim to the skills-walker caller and explicitly document the Copilot asymmetry (symlinks are body-scan-skipped but their target-path strings are archived as `SymbolicLink` entries — not a content-leak, but a documented divergence). Also refreshed the inline symlink-skip comment inside `collectInteriorViolations` to match. Kept `archiveDirectory(agentDir)` with default symlink behavior to preserve the existing Copilot agent-tarball shape; flipping it to `skipSymlinks: true` would be a payload-shape change worth its own PR.

src/agents/skills-walker.ts
chrisleekr: _✅ Addressed_, commit `944eff3`

**Corrected the binary-files docstring: `readFile(utf8)` does not throw on binary content; the regex still runs.**

Spot-on — `readFile(path, "utf8")` substitutes U+FFFD for invalid UTF-8 sequences and returns a string, so binary content IS scanned. Updated the docstring at `src/agents/skills-walker.ts` to reflect reality: only permission errors and transient I/O skip the body scan; binary blobs are scanned and the embedded-secret regex prefix specificity (`sk-ant-api03-`, `AKIA[0-9A-Z]{16}`, `AIza[0-9A-Za-z_-]{35}`, …) keeps false-positive rate near zero while still catching credentials embedded inside binary files. Added an explicit "Do NOT add a binary-detection early-return without a measured false-positive case; it would only weaken coverage." line so a future maintainer cannot misread the comment as license to short-circuit.

chrisleekr:

🔍 **Code review complete**, 7 files, +430/-31.

## Summary

PR #65 closes the bundle-internals literal-secret gap by adding a per-file `scanForSecrets` walk inside the new `collectInteriorViolations` helper, integrating it into both the skills walker and the Copilot agents walker, and broadening `performPush`'s snapshot-warning escalation to a narrow `Detected literal secret (` prefix that catches walker hits without re-catching the redactor warnings already caught at the per-artifact loop. The core walker change, the Copilot agents-loop integration (now consumes both `neverSyncHits` and `secretWarnings` after `eb5e2f1`), the narrowed push-side prefix (`944eff3`), and the test coverage are well-scoped and correct. Earlier review threads (Copilot agents loop dropping `neverSyncHits`, redactor-warning double-report on push abort, and symlink/binary-files docstring inaccuracies) have all been addressed by commits on this branch. I found no remaining behavioral bugs — only two minor nits worth a follow-up. **Verdict: substantively approved; the nits are polish.**

## What was checked

- `git fetch origin main && git log --oneline origin/main..HEAD` — branch is 3 ahead, 0 behind; no rebase needed.
- `git diff --stat origin/main...HEAD` — 7 files, +430/-31.
- Read in full:
  - `src/agents/skills-walker.ts` (new `InteriorViolations`/`collectInteriorViolations`; refreshed gate-4 invocation in `collectSkillArtifacts`; docstring symmetry+binary-files commentary)
  - `src/agents/copilot.ts` (agents-loop integration: consumes BOTH `neverSyncHits` and `secretWarnings`; emits `never-sync inside skill: ` for neverSyncHits to reuse the existing escalation prefix)
  - `src/commands/push.ts` (narrowed snapshot-warning prefix to `Detected literal secret (` and refreshed `.tar.age` skip rationale)
  - `src/agents/__tests__/skills-walker.test.ts` (4 new tests: positive SKILL.md, nested-leak, AKIA-shaped-prose negative, mixed clean+dirty)
  - `src/agents/__tests__/copilot.test.ts` (3 new tests: literal-key agent dir rejected, `auth.json` never-sync hit rejected, clean agent dir → 1 artifact)
  - `src/commands/__tests__/push.test.ts` (end-to-end abort on literal key in `SKILL.md`)
  - `src/commands/__tests__/integration.test.ts` (new double-report regression test)
- Cross-referenced: `collectInteriorViolations` / `collectNeverSyncHits` callers (3 files; helper is now correctly used in both the skills walker and the Copilot agents loop, with no stale `collectNeverSyncHits` references); `Detected literal secret` producers/consumers (`scanForSecrets` and `redactSecretLiterals` in `src/core/sanitizer.ts`); `archiveDirectory` callers (`src/core/tar.ts` `skipSymlinks` semantics); `NEVER_SYNC_PATTERNS` membership; `collect()` helper in `src/agents/_utils.ts`.
- Compared `claude.ts`/`cursor.ts`/`codex.ts` to confirm the Copilot agents loop is the only `archiveDirectory` consumer outside the skills walker — other agents emit per-file `*.md.age` / `*.json.age` artifacts that already flow through the central Phase-1 `scanForSecrets`.
- Validation:
  - `bun install` — 38 packages.
  - `bun run typecheck` — clean, no errors.
  - `bun test src/agents/__tests__/skills-walker.test.ts src/agents/__tests__/copilot.test.ts src/commands/__tests__/push.test.ts` — **46 pass / 0 fail / 136 expect()**.
  - `bun test src/commands/__tests__/integration.test.ts` — **28 pass / 0 fail / 97 expect()**.

## Findings

### [nit] `src/commands/__tests__/integration.test.ts:716` — unanchored `/1 security issue/` regex

The test that locks the abort banner at exactly 1 distinct issue uses `expect(result.errors[0]).toMatch(/1 security issue/)`. The substring `"1 security issue"` is contained in `"Push aborted: 11 security issue(s) detected"` and `"21 security issue(s)"`, so a regression that, say, double-counted twice (4 entries → still 1×4 = 4, no — but 11 is reachable by other shape changes) could falsely pass. The accompanying `expect(literalSecretEntries.length).toBe(1)` already locks the count exactly; the banner assertion should match that strictness.

**Recommended fix:** Replace with `expect(result.errors[0]).toContain("1 security issue(s)")` or anchor with `\b1 security issue\b`. Inline comment posted at `src/commands/__tests__/integration.test.ts:716`.

### [nit] `src/agents/copilot.ts:106` — contract-marker prefix `never-sync inside skill: ` is duplicated as a hardcoded literal across three files without a shared constant

`skills-walker.ts:49` declares `NEVER_SYNC_WARNING_PREFIX` but does not export it; `copilot.ts:106` and `push.ts:151` each repeat the literal verbatim. Same shape applies to the new `Detected literal secret (` prefix in `push.ts:151`. The strings are a load-bearing contract between emitter(s) and the push gate — if any single site drifts (e.g. someone renames the prefix to match the broader semantics now that it covers agent dirs too), the gate silently degrades to non-fatal warnings without a test failure, because each `startsWith` site would independently miss.

**Recommended fix:** Export `NEVER_SYNC_WARNING_PREFIX` and a sibling `WALKER_SECRET_PREFIX = "Detected literal secret ("` from `src/agents/skills-walker.ts` and import them at both consumer sites. Inline comment posted at `src/agents/copilot.ts:106`. Low priority — not a behavioral bug today.

## Reasoning

Several aspects of the diff looked sketchy at first read but check out on inspection:

- **Narrow `Detected literal secret (` prefix at push.ts:151 vs. broad `Detected literal secret` prefix at push.ts:104.** The narrow site only catches walker emissions (`scanForSecrets` always writes `(${name}) in …`); the broad site catches the per-artifact redactor emissions (`Detected literal secret for field …`). The walker writes its warnings to `snapshot.warnings`, not to `artifact.warnings` (skills-walker.ts:242 always sets `warnings: []`). Result: each warning is escalated exactly once. The new `integration.test.ts` regression test seeds the same redactor warning on BOTH arrays and asserts the de-duplication holds — that test would fail if the narrow prefix were widened.
- **scanForSecrets false-positives on hash digests / lockfile blobs / AKIA-shaped prose.** `EMBEDDED_SECRET_PATTERNS` (`src/core/sanitizer.ts:74-83`) are anchored on real credential-prefix tokens with realistic length floors (`AKIA[0-9A-Z]{16}`, `AIza[0-9A-Za-z_-]{35}`, `sk-ant-api03-[A-Za-z0-9_-]{40,}`, etc.). A `sha256:` digest of hex chars and `AKIA-history-of-aviation` (trailing `-` breaks `[0-9A-Z]`) cannot match. The new "does NOT false-positive on a clean skill" test exercises both.
- **Phase-1 chokepoint deliberately skips `.tar.age`.** The rationale (base64 scrambles credential prefixes and overlaps `AKIA…`/`AIza…`) is correct, and the refreshed comment in `push.ts:108-128` now points readers at the walker-layer coverage. The two-layer model is internally consistent.
- **Walker still ships the clean skill in mixed runs.** Confirmed by the "Row 12" test and the new "collects clean skills even when another skill has a literal-secret hit" test. The additive contract is preserved.
- **Symlink asymmetry between skills walker and Copilot agents walker.** The contract docstring at `skills-walker.ts:264-274` explicitly calls this out: skills walker passes `{ skipSymlinks: true }` to `archiveDirectory`, Copilot agents walker uses the default. node-tar archives symlinks as `SymbolicLink` entries (target path only, no content), so this is not a credential-content leak — and the documented divergence is consistent with the prior review's "documented divergence" framing.
- **Binary files getting silently mis-scanned.** Docstring at `skills-walker.ts:276-285` correctly notes that `readFile(path, "utf8")` does NOT throw on binary content — Node replaces invalid sequences with U+FFFD and returns a string; the regex still runs. The high prefix-specificity of `EMBEDDED_SECRET_PATTERNS` keeps false positives near zero while still catching credentials pasted into binary blobs. The explicit "Do NOT add a binary-detection early-return" admonition addresses the concern from the prior review thread.
- **The `never-sync inside skill: ` prefix on an agent dir is semantically inaccurate but pragmatically right.** The existing `push.ts:151` gate matches that exact prefix, and a rename would require coordinated changes to both call sites with no behavioral benefit. The prior review thread accepted this framing.
- **Pre-existing top-level dot-prefix skip asymmetry on Copilot agents loop.** The skills walker skips dot-prefixed top-level entries (gate 1 at skills-walker.ts:172); the Copilot agents loop at copilot.ts:97-101 does not. Pre-existing — the diff for this PR is purely additive on this loop. Out of scope.
- **`scanForSecrets` reading every readable interior file allocates memory proportional to file size.** Skill/agent dirs are user-authored configs (small markdown/json files); symlinks are skipped so a vendored multi-GB blob symlinked into a skill cannot drag the walker to a halt. Not a real performance concern in this PR's scope.

The single deliberate UX regression the prior review flagged (duplicate-escalation on push.ts:144) was addressed in commit `944eff3`. The Copilot agents-loop omission flagged in the resolve thread was addressed in commit `eb5e2f1`. Both fixes have regression tests on this branch. Nothing else stands out.

_cost: $3.3113 · turns: 34 · duration: 330s_

src/commands/__tests__/integration.test.ts
```diff
@@ -675,6 +675,58 @@ describe("integration", () => {
...
+    pushMod.__setPushAgentsForTesting(dualSnapshotAgent);
+    try {
+      const result = await pushMod.performPush({ agent: "claude" });
+
+      expect(result.fatal).toBe(true);
+      expect(result.pushed).toBe(0);
+      const literalSecretEntries = result.errors.filter((e) => e.includes(duplicatedWarning));
+      expect(literalSecretEntries.length).toBe(1);
+      // The banner's count must match the distinct-issue count, not 2× it.
+      expect(result.errors[0]).toMatch(/1 security issue/);
```
chrisleekr: _💭 Suggestion_      | _🔵 Nit_

**Unanchored `/1 security issue/` regex would silently pass for any count ending in `1`.**

The regex `/1 security issue/` matches the substring anywhere in the banner. `"Push aborted: 11 security issue(s) detected"` and `"Push aborted: 21 security issue(s) detected"` both contain `"1 security issue"`, so the test would falsely pass if a future regression bumped the count from 1 to 11 (or 21, 31…). For a test whose purpose is to lock the count at exactly 1, the assertion should be exact.

**Suggested fix:** Replace with `expect(result.errors[0]).toContain("1 security issue(s)")` (or anchor the regex with `\b`) so any non-`1` count fails. The companion `expect(literalSecretEntries.length).toBe(1)` above already locks the entry count exactly — this banner assertion should match that strictness.

src/agents/copilot.ts
```diff
@@ -77,13 +82,33 @@ export async function snapshotCopilot(): Promise<SnapshotResult> {
...
     const names = await readdir(AgentPaths.copilot.agentsDir);
     for (const name of names) {
       const agentDir = join(AgentPaths.copilot.agentsDir, name);
       const agentDirStat = await stat(agentDir).catch(() => null);
       if (!agentDirStat?.isDirectory()) continue;
+
+      const violations = await collectInteriorViolations(agentDir);
+      if (violations.neverSyncHits.length > 0 || violations.secretWarnings.length > 0) {
+        for (const hit of violations.neverSyncHits) {
+          warnings.push(`never-sync inside skill: ${hit}`);
```
chrisleekr: _💭 Suggestion_      | _🔵 Nit_

**Contract-marker prefix `never-sync inside skill: ` is now a hardcoded literal in three places without a shared constant.**

`skills-walker.ts:49` declares `NEVER_SYNC_WARNING_PREFIX` but does not export it; `copilot.ts:106` and `push.ts:151` each repeat the literal verbatim. The string is a load-bearing contract between the walker (emitter), the agents adapter (emitter for the agent-dir path), and the push gate (escalator) — if any single site drifts (e.g. someone renames it to `never-sync inside bundle: ` to match the now-broader semantics), the gate silently degrades to non-fatal warnings without a test failure, because the per-prefix `startsWith` test would just stop matching. Same shape applies to `Detected literal secret (` at push.ts:151.

**Suggested fix:** Export `NEVER_SYNC_WARNING_PREFIX` (and a sibling `WALKER_SECRET_PREFIX = "Detected literal secret ("`) from `src/agents/skills-walker.ts` and import them at both consumer sites. Keeps the three sites in lockstep and makes the contract greppable. Low-priority — not a behavioral bug today.

chrisleekr:

## Outstanding

Nothing blocks merge from a checks/comments standpoint. Open items are owner actions:

- **No formal approval yet.** `reviewDecision: null` — the human reviewer (@chrisleekr) needs to approve the PR before GitHub allows merge.
- **CodeRabbit review still running** (optional, non-blocking). `mergeStateStatus: UNSTABLE` will flip to `CLEAN` once it finishes.

_cost: $3.0455 · turns: 7 · duration: 65s_

src/commands/__tests__/integration.test.ts
chrisleekr: _✅ Addressed_, commit `c2bd1fd`

**Tightened banner assertion to an exact substring match.**

You're right — `/1 security issue/` is a loose substring match that silently passes for `11 security issue(s)`, `21 security issue(s)`, etc., so a future regression that bumped the dedupe count from 1 to 11 (or any number ending in `1`) would slip through. Replaced with `expect(result.errors[0]).toContain("1 security issue(s)")` at `src/commands/__tests__/integration.test.ts:716` so the banner assertion now locks the count as strictly as the companion `expect(literalSecretEntries.length).toBe(1)` above it.

src/agents/copilot.ts
chrisleekr: _✅ Addressed_, commit `c2bd1fd`

**Promoted the contract prefixes to exported constants.**

You're right — three independent `startsWith` / template-literal sites for two load-bearing strings was a silent-drift hazard. Exported `NEVER_SYNC_WARNING_PREFIX = "never-sync inside skill: "` (already declared, now exported) and added `WALKER_SECRET_WARNING_PREFIX = "Detected literal secret ("` from `src/agents/skills-walker.ts`. Both consumer sites (`src/agents/copilot.ts` emit, `src/commands/push.ts` escalate-gate) now import them. If any future maintainer wants to rename either prefix they have to change the constant once and the three call sites line up automatically — no chance the escalation gate silently degrades to non-fatal warnings because one `startsWith` literal drifted from the emitter.

PR #132, opened 2026-05-21
fix(daemon): use os.tmpdir() for Windows task XML path

### LINKED ISSUE #125: fix(cross-platform): Windows daemon installer falls back to non-existent C:\Temp instead of os.tmpdir()
**Finding**

`src/daemon/installer-windows.ts:159` resolves a temp file path for the
Task Scheduler XML by hand. The line interpolates `process.env.TEMP`
(falling back to the literal string `C:\Temp` when it is undefined) and
appends `\agentsync-task.xml`.

This is the only place in the codebase that resolves a temp directory by
reading `process.env.TEMP` directly. Both `src/commands/daemon.ts:2,19`
and `src/core/version-check.ts:3,51` already use `tmpdir()` from
`node:os`. The Windows installer is the lone outlier, and it has two
cross-platform-specific defects:

1. The fallback `C:\Temp` is not a standard Windows directory. Stock
   Windows has no `C:\Temp`. The OS-managed temp locations are
   `%TEMP%` / `%TMP%` (per-user) and `%SystemRoot%\Temp` (i.e.
   `C:\Windows\Temp`). When `TEMP` is unset and `C:\Temp` does not exist,
   `writeFile(tmpXml, xml, "utf16le")` at line 162 throws `ENOENT` and
   `installWindows()` aborts before `schtasks /Create` ever runs — so
   `agentsync daemon install` fails on Windows with an opaque filesystem
   error instead of installing the daemon.

2. The nullish-coalescing fallback does not treat a blank value as
   unset. It substitutes the fallback only when `TEMP` is `undefined`.
   An exported-but-empty `TEMP=` yields a tmpXml of `\agentsync-task.xml`,
   which writes to the root of the current drive. This is exactly the
   class of bug the project has been fixing repeatedly elsewhere —
   `nonBlank()` from `src/lib/env.ts` is already applied to `APPDATA`,
   `CODEX_HOME`, and `AGENTSYNC_DIR` in `src/config/paths.ts`, and recent
   commits 117, 114, 111 hardened other env-var reads against blank
   values. The installer was never brought in line.

`os.tmpdir()` solves both: on Windows it checks `TEMP`, then `TMP`, then
`%SystemRoot%\temp` / `%windir%\temp` — a directory that always exists on
a normal install — and treats empty strings as unset.

**Diagram**

```mermaid
graph TD
    Start["agentsync daemon install on Windows"] --> Cur["installer-windows.ts line 159<br/>process.env.TEMP fallback C-Temp"]
    Cur --> Q1["Is TEMP set and non-empty?"]
    Q1 -->|yes| OK1["writes to real temp dir — works"]
    Q1 -->|undefined| Fallback["tmpXml = C-Temp + agentsync-task.xml"]
    Q1 -->|empty string| Root["tmpXml = drive-root + agentsync-task.xml"]
    Fallback --> Q2["Does C-Temp exist?"]
    Q2 -->|no — stock Windows| Fail["writeFile ENOENT<br/>daemon install aborts"]:::bad
    Q2 -->|yes| OK2["works by luck"]
    Root --> Risk["wrong location or permission error"]:::bad
    Cur -.proposed.-> Fix["join of os.tmpdir and agentsync-task.xml<br/>TEMP then TMP then SystemRoot Temp"]:::good
    Fix --> Always["always resolves an existing dir"]:::good

    classDef bad fill:#c0392b;color:#ffffff;stroke:#7b241c
    classDef good fill:#1e8449;color:#ffffff;stroke:#196f3d
```

**Rationale**

Windows daemon support is a first-class, advertised feature, and the
daemon installer is on the critical path for it — yet it depends on a
directory (`C:\Temp`) that does not exist on a default Windows install.
The failure mode is silent and total: install aborts with a raw `ENOENT`
and no guidance, on the exact platform the file targets.

The fix is tiny — one import plus one line — carries no behavior change
for the common case (`TEMP` set), and aligns the installer with the
`tmpdir()` pattern the rest of the codebase already uses. It also closes
the blank-env-var gap consistently with the established `nonBlank()`
usage already present in `paths.ts`.

**References**

Internal:
- `src/daemon/installer-windows.ts:159` — temp path built from `process.env.TEMP` with a `C:\Temp` fallback
- `src/daemon/installer-windows.ts:161-162` — `writeFile(tmpXml, ...)` that throws on a missing dir
- `src/daemon/installer-windows.ts:155-181` — `installWindows()` aborts before `schtasks /Create`
- `src/commands/daemon.ts:2,19` — existing `tmpdir()` from `node:os` usage
- `src/core/version-check.ts:3,51` — existing `tmpdir()` from `node:os` usage
- `src/config/paths.ts` — `nonBlank()` already applied to `APPDATA`, `CODEX_HOME`, `AGENTSYNC_DIR`
- `src/lib/env.ts` — `nonBlank()` helper

External:
- Node.js OS API documentation, os.tmpdir(): https://nodejs.org/api/os.html
- nodejs/node PR 42300, temp env-var resolution order on Windows: https://github.com/nodejs/node/pull/42300
- nodejs/node issue 60582, os.tmpdir edge cases on Windows: https://github.com/nodejs/node/issues/60582

**Suggested Next Steps**

1. In `src/daemon/installer-windows.ts`, import `tmpdir` from `node:os`
   and `join` from `node:path`, then replace the line 159 template
   literal so tmpXml becomes `join(tmpdir(), "agentsync-task.xml")`.
2. Add a test in `src/daemon/__tests__/installer-windows.test.ts` that
   asserts `installWindows()` writes the XML under `os.tmpdir()` (the
   existing `fs` injection slot makes the `writeFile` path observable),
   and a case covering an empty `TEMP` to lock in the blank-env-var behavior.

**Areas Evaluated**

- `src/config/paths.ts` — path resolution, `APPDATA` / `CODEX_HOME` / `AGENTSYNC_DIR`
  handling, Unix socket vs Windows named-pipe endpoint. `nonBlank()` correctly
  applied to overrides; no finding beyond the inconsistency noted above.
- `src/daemon/installer-windows.ts` — Task Scheduler XML build, `schtasks`
  invocation, arg quoting. Finding here.
- `src/daemon/installer-linux.ts` — systemd user unit; uses `homedir()` and
  `join()` throughout, no temp-file dependency. No finding.
- `src/daemon/installer-macos.ts` — launchd plist; uses `homedir()` and
  `join()` throughout, no temp-file dependency. No finding.
- `src/core/ipc.ts` — Unix-socket / named-pipe IPC; stale-socket unlink is
  guarded for both platforms. No finding.

---
*Generated by scheduled research workflow on 2026-05-18*

chrisleekr:
## Summary

Fix the Windows daemon installer so `agentsync daemon install` does not
abort with `ENOENT` on a stock Windows install. The previous code hand-
rolled the temp path as `` `${process.env.TEMP ?? "C:\\Temp"}\\agentsync-task.xml` ``,
which has two cross-platform defects: (1) the fallback `C:\Temp` does not
exist on stock Windows, so `writeFile` throws and the installer aborts
before `schtasks /Create` runs; (2) `??` does not treat an exported-but-
empty `TEMP=` as unset, so the path collapses to `\agentsync-task.xml`
at drive root. Replace with `join(tmpdir(), "agentsync-task.xml")`, which
checks `TEMP` → `TMP` → `%SystemRoot%\Temp` and treats blanks as unset —
matching the existing pattern in `src/commands/daemon.ts` and
`src/core/version-check.ts`.

## Changes

- `src/daemon/installer-windows.ts`: import `tmpdir` from `node:os` and
  `join` from `node:path`; rebuild `tmpXml` in `installWindows()` via
  `join(tmpdir(), "agentsync-task.xml")`. No behaviour change when
  `TEMP` is set and non-empty.
- `src/daemon/__tests__/installer-windows.test.ts`: add two regression
  tests against the existing `fsStub` injection: one asserts the write
  path lives under `os.tmpdir()` and ends with `agentsync-task.xml`;
  the other sets `TEMP=""` and asserts the path does not collapse to
  drive root. The blank-`TEMP` test saves and restores `process.env.TEMP`
  in a `try/finally`.

## Diagram

```mermaid
graph TD
    A["agentsync daemon install (Windows)"] --> B["installWindows(args)"]
    B --> C{"Resolve tmpXml"}
    C -->|before| D1["process.env.TEMP ?? C:\\Temp + \\agentsync-task.xml"]
    D1 --> E1{"TEMP state"}
    E1 -->|set, non-empty| F1["writes to real temp"]
    E1 -->|undefined| G1["C:\\Temp\\agentsync-task.xml"]
    G1 --> H1["writeFile ENOENT on stock Windows — install aborts"]:::bad
    E1 -->|empty string| I1["\\agentsync-task.xml at drive root"]:::bad
    C -->|after| D2["join(tmpdir(), agentsync-task.xml)"]
    D2 --> F2["tmpdir() checks TEMP then TMP then SystemRoot Temp; blanks treated as unset"]:::good
    F2 --> J["writeFile succeeds → schtasks /Create runs"]:::good
    classDef bad fill:#c0392b,color:#ffffff,stroke:#7b241c
    classDef good fill:#1e8449,color:#ffffff,stroke:#196f3d
```

## Files changed (path · one-line rationale)

- `src/daemon/installer-windows.ts` · swap hand-rolled temp path for `join(tmpdir(), "agentsync-task.xml")`; add `node:os` and `node:path` imports.
- `src/daemon/__tests__/installer-windows.test.ts` · add two regression tests pinning the `tmpdir()` invariant (default + blank-`TEMP`).

## Commits (sha · subject)

- `a718251` · fix(daemon): use os.tmpdir() for Windows task XML path

## Tests run (command · result)

- `bun test src/daemon/__tests__/installer-windows.test.ts` · pass — 18 pass, 0 fail (16 pre-existing + 2 new regression guards).
- `bun run typecheck` · pass (`tsc --noEmit`).
- `bun run lint` · pass (Biome, 14 pre-existing warnings, 0 errors — none introduced by this PR).
- `bun run check:spec-refs` · pass — `spec-tracker leakage check passed (139 files)`.
- `bun test` (full suite) · 1 pre-existing failure in `src/commands/__tests__/daemon.test.ts > install subcommand calls installer.install with args array`. Reproduced on `main` from the same sandbox before any changes: the workspace lives under `/tmp`, which trips `isEphemeralPath()` in `src/commands/daemon.ts:33-36`. Not caused by this PR — no daemon-command code is touched.

## Verification

- `rg "process\.env\.TEMP" src` returns no results — the installer was the lone direct reader, now removed.
- `src/daemon/installer-windows.ts` imports `tmpdir` from `node:os` and `join` from `node:path`; `tmpXml` is built via `join(tmpdir(), "agentsync-task.xml")`.
- New test `writes XML under os.tmpdir() with the agentsync-task.xml filename` asserts the resolved path satisfies `xmlPath.startsWith(tmpdir())` and ends with `agentsync-task.xml`.
- New test `resolves under os.tmpdir() even when TEMP is set to empty string` sets `process.env.TEMP = ""`, invokes `installWindows`, asserts the path is still under `tmpdir()` and is not the drive-root `\agentsync-task.xml`, then restores the env in `finally`.

Closes #125

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Windows installer reliability by fixing temporary task file path resolution to ensure correct system temporary directory usage regardless of environment configuration.

* **Tests**
  * Added regression tests to verify scheduled task file placement in the system temporary directory and confirm proper handling of misconfigured environment variables.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/132?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/daemon/__tests__/installer-windows.test.ts
+++ b/src/daemon/__tests__/installer-windows.test.ts
@@ -8,7 +8,9 @@
  * (`__setInstallerWindowsImplForTests`). Bypasses Bun's mock.module()
  * entirely so daemon.test.ts's installer overrides cannot bleed in.
  */
+
 import { afterAll, beforeAll, beforeEach, describe, expect, test } from "bun:test";
+import { tmpdir } from "node:os";
 
 const fsWrites = new Map<string, string>();
 const fsRms = new Set<string>();
@@ -150,6 +152,40 @@ describe("installWindows", () => {
 
     const xmlEntry = [...fsRms].find((p) => p.endsWith(".xml"));
     expect(xmlEntry).toBeDefined();
+  });
+
+  // Regression guard: the install path must always resolve under os.tmpdir().
+  // Hand-rolling `process.env.TEMP ?? "C:\\Temp"` previously produced a path
+  // whose parent does not exist on stock Windows, aborting installWindows()
+  // with ENOENT before schtasks /Create ever ran.
+  test("writes XML under os.tmpdir() with the agentsync-task.xml filename", async () => {
+    await m.installWindows(["C:\\agentsync.exe"]);
+
+    const xmlPath = [...fsWrites.keys()].find((p) => p.endsWith("agentsync-task.xml"));
+    expect(xmlPath).toBeDefined();
+    if (!xmlPath) return;
+    expect(xmlPath.startsWith(tmpdir())).toBe(true);
+  });
+
+  // Regression guard: an exported-but-empty TEMP env var must not collapse
+  // the write path to drive root. `??` would short-circuit only on undefined,
+  // letting "" through; os.tmpdir() treats blanks as unset and falls through
+  // its own resolution chain.
+  test("resolves under os.tmpdir() even when TEMP is set to empty string", async () => {
+    const saved = process.env.TEMP;
+    process.env.TEMP = "";
+    try {
+      await m.installWindows(["C:\\agentsync.exe"]);
+
+      const xmlPath = [...fsWrites.keys()].find((p) => p.endsWith("agentsync-task.xml"));
+      expect(xmlPath).toBeDefined();
+      if (!xmlPath) return;
+      expect(xmlPath.startsWith(tmpdir())).toBe(true);
+      // Defence-in-depth: must not be the drive-root regression
+      expect(xmlPath).not.toBe("\\agentsync-task.xml");
+    } finally {
+      process.env.TEMP = saved;
+    }
   });
 });
 
--- a/src/daemon/installer-windows.ts
+++ b/src/daemon/installer-windows.ts
@@ -8,6 +8,8 @@
  */
 import { execFile } from "node:child_process";
 import type * as fsPromises from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { promisify } from "node:util";
 import { log } from "@clack/prompts";
 
@@ -156,7 +158,9 @@ export async function installWindows(args: string[]): Promise<void> {
   const override = getOverride("installWindows");
   if (override) return override(args);
   const xml = buildXml(args);
-  const tmpXml = `${process.env.TEMP ?? "C:\\Temp"}\\agentsync-task.xml`;
+  // Use os.tmpdir() so the resolved directory always exists: on Windows it
+  // checks TEMP → TMP → %SystemRoot%\Temp and treats blank values as unset.
+  const tmpXml = join(tmpdir(), "agentsync-task.xml");
 
   const { writeFile, rm } = await fsImpl();
   await writeFile(tmpXml, xml, "utf16le");


chrisleekr:

🔍 **Code review complete**, 2 files, +41/-1.

## Summary

PR #132 replaces a hand-rolled Windows temp path (`process.env.TEMP ?? "C:\\Temp"`) in `installWindows()` with `join(tmpdir(), "agentsync-task.xml")`, and adds two regression tests. The fix is correct, minimal, well-scoped, and consistent with existing usage of `node:os` `tmpdir()` in `src/commands/daemon.ts:19`, `src/core/version-check.ts:51`, and several test files. Typecheck passes, lint shows only the 14 pre-existing warnings, and the new test file passes 18/18. No findings worth blocking on; recommended to merge.

## What was checked

Files read in full:
- `src/daemon/installer-windows.ts` (entire file, not just the diff)
- `src/daemon/__tests__/installer-windows.test.ts` (entire file)

Cross-references performed:
- `rg "tmpdir|process\\.env\\.TEMP|process\\.env\\.TMP" src` to confirm the rest of the codebase uses `node:os` `tmpdir()` and that no other caller still hand-rolls `process.env.TEMP`. The old pattern was indeed unique to `installer-windows.ts`.
- `src/commands/daemon.ts:19` and `src/core/version-check.ts:51` are the analogous existing callers cited in the PR description; both use `tmpdir()` directly, so this PR brings the installer in line.
- Read the surrounding `installWindows()` body (lines 157-185) to confirm the `tmpXml` variable is the only consumer of the resolved path, and that the existing `try/finally` cleanup around `schtasks /Create` is unchanged.

Validation runs:
- `bun test src/daemon/__tests__/installer-windows.test.ts` -> 18 pass, 0 fail (16 pre-existing + 2 new guards).
- `bun run typecheck` -> pass (clean `tsc --noEmit`).
- `bun run lint` -> 14 pre-existing warnings, 0 errors; none introduced by this PR.
- Empirically verified the technical claim that `os.tmpdir()` treats blank env vars as unset by running `TEMP="" TMP="" TMPDIR="" bun -e '...'` and observing it falls back to `/tmp`. This matches libuv's modern `CHECK_ENV_VAR` macro, which short-circuits only when the env var is non-NULL **and** non-empty. The PR's stated rationale ("treats blanks as unset") is accurate on both POSIX and Windows.

## Findings

No findings, see Reasoning.

## Reasoning

Things that initially looked sketchy but check out on inspection:

1. **First new test ("writes XML under os.tmpdir() with the agentsync-task.xml filename") is a weak regression guard against the original bug.** Because `beforeAll` sets `process.env.TEMP = "/tmp/agent-sync-test"` and both the implementation and the assertion call `tmpdir()`, the test would also pass against the old `process.env.TEMP ?? "C:\\Temp"` code path on Linux (the resolved path would still start with `/tmp/agent-sync-test`, just with a literal backslash separator). The second test (the `TEMP=""` case) is what actually catches the `??` vs blank-fallback bug via the `not.toBe("\\agentsync-task.xml")` defence-in-depth assertion. I considered flagging the first test as Minor, but decided against it: it still pins the directory contract and the filename, and the blank-TEMP test does the heavy lifting. Calling out a weak-but-not-wrong assertion would be thoroughness theatre.

2. **`originalTemp` could be `undefined` at module load, causing `process.env.TEMP = undefined` to set TEMP to the literal string `"undefined"` in `afterAll`.** This is a real Node.js gotcha, but it's a pre-existing pattern in the file (lines 54, 63) and not introduced by this PR. The new test's `try/finally` saves `saved = process.env.TEMP` only **after** `beforeAll` has set it to `/tmp/agent-sync-test`, so `saved` is always a string and the restore is safe within the new test's scope. Not in scope for this PR.

3. **The `expect(xmlPath).not.toBe("\\agentsync-task.xml")` assertion uses a literal Windows-style backslash on a test that runs on Linux.** This is correct: the assertion is checking that the resolved path is *not* the buggy old-code output. On Linux the path would never be that literal anyway, so the assertion is functionally a no-op on POSIX, but it documents intent and would catch a regression on a Windows CI runner. Worth keeping.

4. **The fsStub mocks `writeFile`, so the actual ENOENT failure on stock Windows can't be reproduced in CI on Linux.** This is inherent to cross-platform testing, not a defect of this PR. The fix is verified by the path-resolution assertions; the I/O failure mode is sufficiently described in the PR body and the inline comment in `installer-windows.ts:161-162`.

5. **No callers of `installWindows()` outside of `src/commands/daemon.ts` (via the `PlatformInstaller` interface) and the test file.** Confirmed by reading `src/commands/daemon.ts:84-96`. The signature is unchanged (`(args: string[]) => Promise<void>`), so the contract is preserved.

Verdict: clean, minimal, correct fix with adequate regression coverage. Recommend merge.

_cost: $1.4773 · turns: 18 · duration: 280s_

chrisleekr:

## Outstanding

Nothing blocking merge from the resolve agent's side:

- CI is fully green on `a718251`.
- No open review threads.
- Branch is current with `main`.
- Maintainer has signalled merge.

The only nit worth surfacing: `gh pr view --json reviewDecision` returns
empty (no formal GitHub PR review submitted yet — the maintainer's
"merge" signal came through the digest, not via `gh pr review --approve`).
Per instructions, the "resolve complete, ready to merge" one-liner is
gated on `reviewDecision === APPROVED`, so this run stops at the summary
rather than posting that line. Merge itself is a human action (no
`gh pr merge` call from this agent).

_cost: $0.4112 · turns: 6 · duration: 57s_

PR #133, opened 2026-05-21
fix(config): apply nonBlank to Windows APPDATA in all agent paths

### LINKED ISSUE #126: fix(config): Windows APPDATA agent-path resolvers missed by #117 blank-env fix, collapse to relative paths
## Finding

`src/config/paths.ts` resolves three Windows agent-config paths from `%APPDATA%` using the bare pattern `process.env.APPDATA ?? ""`:

- `src/config/paths.ts:23` — `cursor.settingsJson`: `join(process.env.APPDATA ?? "", "Cursor", "User", "settings.json")`
- `src/config/paths.ts:85` — `copilot.vscodeMcpInSettings`: `join(process.env.APPDATA ?? "", "Code", "User", "settings.json")`
- `src/config/paths.ts:96` — `vscode.mcpJson`: `join(process.env.APPDATA ?? "", "Code", "User", "mcp.json")`

This is the exact bug class that commit `0598a11` ("fix(config): treat blank APPDATA and CODEX_HOME env vars as unset", #117) fixed — but that fix only touched `resolveAgentSyncHome` (`paths.ts:149`, now `nonBlank(process.env.APPDATA) ?? HOME`) and `codex` `CODEX_HOME` (`paths.ts:49`). The three agent-path resolvers were left on the old pattern, so `paths.ts` is now internally inconsistent: `nonBlank` is imported and used in two places, ignored in three others.

Two distinct failure modes follow:

1. **Blank `APPDATA` (`APPDATA=""`)** — an exported-but-empty env var is `""`, which is *not* nullish, so `?? ""` does not trigger. `join("", "Cursor", "User", "settings.json")` yields the relative path `Cursor\User\settings.json`. The `nonBlank` helper (`src/lib/env.ts`) exists precisely to collapse this case to `undefined`.
2. **Unset `APPDATA`** — `?? ""` triggers, but the fallback is `""`, again producing the relative path `Cursor\User\settings.json`. There is no sensible default; `resolveAgentSyncHome` at least falls back to `HOME`.

Either way, on Windows the resolved path is **relative**, so it resolves against `process.cwd()` of whatever process touches it (CLI invocation dir, or the daemon's working directory):

- **Push / snapshot**: the walker stats a non-existent relative path → the file is treated as absent → Cursor settings, VS Code MCP config, and Copilot's VS Code settings are **silently dropped from the vault** with no error.
- **Pull / apply**: the decrypted artefact is written to `<cwd>\Cursor\User\settings.json` — a wrong, unexpected location — **silently misplacing the user's config** instead of restoring it.

This contradicts the project's "errors over fallbacks" rule (CLAUDE.md) and the architecture doc's claim that "platform-specific decisions live in `src/config/paths.ts`" reliably.

## Diagram

```mermaid
graph TD
    Env["process.env.APPDATA on Windows"] --> Blank{"blank or unset?"}
    Blank -->|"unset"| Fb1["nullish coalesce triggers -> empty"]:::bad
    Blank -->|"blank string"| Fb2["nullish coalesce does NOT trigger -> empty"]:::bad
    Blank -->|"valid value"| Ok["absolute APPDATA path"]:::good
    Fb1 --> Rel["join with empty base = relative path"]:::bad
    Fb2 --> Rel
    Rel --> Push["push: file not found -> config silently skipped"]:::bad
    Rel --> Pull["pull: writes under process cwd -> config misplaced"]:::bad
    Ok --> Good["push and pull hit the real config file"]:::good

    classDef bad fill:#c0392b;color:#ffffff;stroke:#7b241c
    classDef good fill:#1e8449;color:#ffffff;stroke:#196f3d
```

## Rationale

- **Silent data loss / misplacement** is the worst failure category for a sync tool — the user gets no abort, no warning, just a vault missing files or a machine with config written to a junk directory.
- **Scope is real**: it hits every Windows user whose `APPDATA` is unset or blank — a state that occurs in stripped service accounts, CI runners, container images, and `runas`/scheduled-task contexts where the env block is not fully populated. The daemon runs under Task Scheduler, exactly the kind of context where a minimal env block is common.
- **The fix is one-line-per-site** and the helper (`nonBlank`) plus the canonical default already exist in the codebase — this is finishing #117, not new design.
- It removes a genuine inconsistency: a reader of `paths.ts` sees `nonBlank` used for `CODEX_HOME` and `resolveAgentSyncHome` but not for the sibling `APPDATA` lookups, and cannot tell whether that is intentional.

## References

- **Internal**:
  - `src/config/paths.ts:23`, `:85`, `:96` — the three `process.env.APPDATA ?? ""` sites
  - `src/config/paths.ts:49`, `:149` — the already-fixed `nonBlank(...)` sites for comparison
  - `src/lib/env.ts:8` — `nonBlank()` helper
  - commit `0598a11` (#117) — prior fix that missed these three sites
  - `src/config/__tests__/paths.test.ts:136-168` — existing `AGENTSYNC_DIR` blank/whitespace tests show the established test pattern; no equivalent exists for `APPDATA`
  - `CLAUDE.md` — "Errors over fallbacks" convention; `docs/architecture.md` "Path resolution" section
- **External**:
  - [Recognized environment variables — Microsoft Learn](https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables) — `%APPDATA%` = `%USERPROFILE%\AppData\Roaming`
  - [AppData / LocalAppData / ProgramData explanations](https://www.advancedinstaller.com/appdata-localappdata-programdata.html)

## Suggested Next Steps

1. Replace `process.env.APPDATA ?? ""` at `paths.ts:23`, `:85`, `:96` with `nonBlank(process.env.APPDATA) ?? join(HOME, "AppData", "Roaming")` — the canonical default keeps resolution absolute even when `APPDATA` is unset or blank. Consider a small local `const APPDATA = nonBlank(process.env.APPDATA) ?? join(HOME, "AppData", "Roaming")` so all three win32 branches share one definition.
2. Add Windows-guarded tests to `src/config/__tests__/paths.test.ts` asserting that, with `APPDATA` unset and with `APPDATA=""`, the three resolved paths are absolute (`path.isAbsolute(...)` is true) — mirroring the existing `AGENTSYNC_DIR` blank/whitespace tests.
3. Alternative (stricter, matches "errors over fallbacks"): if no absolute `%APPDATA%` can be resolved, surface an actionable error from the snapshot/apply layer rather than writing to a relative path — but the canonical-default approach in step 1 is the lower-risk, behavior-preserving fix and is consistent with how `resolveAgentSyncHome` already handles the same variable.

## Areas Evaluated

- `src/config/schema.ts` — recipient/remote/sync validation; already hardened (see closed issue on empty recipients map / non-`age1` keys). No new finding.
- `src/config/loader.ts` — `loadConfig`/`writeConfig`/`resolveConfigPath`; `structuredClone` Symbol-stripping is sound; TOML round-trip covered by tests. No finding.
- `src/config/paths.ts` — **finding here**: three `APPDATA` resolvers missed by the #117 blank-env fix.
- `src/lib/env.ts` — `nonBlank` helper is correct; the issue is that `paths.ts` does not apply it consistently.
- `src/config/__tests__/` — `schema.test.ts`, `loader.test.ts`, `paths.test.ts` reviewed; no `APPDATA` coverage exists for the win32 branches.

---
*Generated by scheduled research workflow on 2026-05-19*

chrisleekr:
## Summary

Finishes #117 by routing the three Windows agent-config paths in
`src/config/paths.ts` through the same `nonBlank` helper already used
for `CODEX_HOME` and `resolveAgentSyncHome`. The `cursor.settingsJson`,
`copilot.vscodeMcpInSettings`, and `vscode.mcpJson` win32 branches all
resolved `%APPDATA%` via the legacy `process.env.APPDATA ?? ""`
pattern. An exported-but-empty `APPDATA=""` is not nullish, so `??`
did not trigger; when it did trigger, the fallback was `""`. Both
cases produced a relative path, so snapshot silently dropped the
file and apply wrote the decrypted artefact under `process.cwd()`.
This is exactly the failure mode commit 0598a11 (#117) fixed, just at
three sites that were missed.

Closes #126

## Changes

- Introduced a module-level `const APPDATA = nonBlank(process.env.APPDATA) ?? join(HOME, "AppData", "Roaming")` near the existing `HOME` / `PLATFORM` constants in `src/config/paths.ts`, so all three win32 branches share one absolute default.
- Replaced the three legacy `process.env.APPDATA ?? ""` sites at `src/config/paths.ts:23`, `:85`, and `:96` with the shared `APPDATA` constant.
- Added a Windows-guarded `describe` block in `src/config/__tests__/paths.test.ts` modelled on the existing `AGENTSYNC_DIR` blank/whitespace tests. It re-imports `../paths` with a cache-busting query suffix under `APPDATA` unset / `""` / `"   "` and asserts `path.isAbsolute()` for all three affected paths. The assertions short-circuit on non-Windows so Linux/macOS CI stays green.

## Diagram

```mermaid
graph TD
    Env["process.env.APPDATA on Windows"] --> Helper["nonBlank(process.env.APPDATA)"]
    Helper -->|"valid value"| Use["use trimmed APPDATA"]
    Helper -->|"undefined, blank, whitespace"| Fallback["join(HOME, AppData, Roaming)"]
    Use --> Abs["absolute Cursor / Copilot / VS Code config path"]
    Fallback --> Abs
    Abs --> Push["push: stats the real file, captured in vault"]
    Abs --> Pull["pull: writes to the real config location"]
```

## Files changed

- `src/config/paths.ts` · introduce shared `APPDATA` constant via `nonBlank` with `HOME/AppData/Roaming` default; route the three win32 branches through it.
- `src/config/__tests__/paths.test.ts` · add Windows-guarded `describe` block asserting all three resolved paths stay absolute when `APPDATA` is unset, empty, or whitespace-only; import `isAbsolute` from `node:path`.

## Commits

- `75d84de` · fix(config): apply nonBlank to Windows APPDATA in all agent paths

## Tests run

- `bun test src/config/__tests__/paths.test.ts` · result: 26 pass, 0 fail (the 3 new Windows-guarded tests short-circuit on Linux but still run their `beforeEach` / `afterEach` envelope).
- `bun run typecheck` · result: passes (`tsc --noEmit` clean).
- `bun run lint` · result: 14 warnings + 1 info, all pre-existing in other files (`scripts/*.ts`, `src/core/path-portability.ts`, etc.); no new findings in `src/config/paths.ts` or `paths.test.ts`.
- `bun run check` · result: the spec-refs scanner and docs-mirror checks pass; the full suite surfaces one pre-existing failure in `src/commands/__tests__/daemon.test.ts` ("Executable is in a temporary directory") that also fails on `main` and is unrelated to this change.

## Verification

- `grep -n 'process.env.APPDATA' src/config/paths.ts` returns exactly two lines: the new `APPDATA` constant (line 13) and the existing `resolveAgentSyncHome` site (line 156). Both are wrapped by `nonBlank(...)`. No bare `?? ""` pattern remains.
- A reader of `src/config/paths.ts` now sees `nonBlank(process.env.APPDATA)` applied consistently across all four APPDATA sites (the three agent paths plus `resolveAgentSyncHome`), eliminating the previous internal inconsistency.
- On a Windows host with `APPDATA` unset or blank, the three resolved paths are absolute and anchored at `<HOME>\AppData\Roaming\...`, so snapshot no longer silently skips them and apply no longer writes under `process.cwd()`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows path resolution to properly handle cases where the APPDATA environment variable is unset or contains only whitespace, ensuring absolute paths are generated with a fallback to `HOME/AppData/Roaming`.

* **Tests**
  * Added test coverage for Windows APPDATA fallback scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chrisleekr/agentsync/pull/133?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


chrisleekr:

🔍 **Code review complete**, 2 files, +51/-4.

## Summary

Tight, well-scoped follow-up to #117. The PR eliminates the last three `process.env.APPDATA ?? ""` sites in `src/config/paths.ts` (cursor `settingsJson`, copilot `vscodeMcpInSettings`, vscode `mcpJson`) by routing them through a shared module-level constant `APPDATA = nonBlank(process.env.APPDATA) ?? join(HOME, "AppData", "Roaming")`. The fix is correct, the explanatory comment is accurate, and there is no behavioral change on darwin/linux. **No findings** — recommend merge.

## What was checked

- Read both changed files in full: `src/config/paths.ts` and `src/config/__tests__/paths.test.ts`.
- Read `src/lib/env.ts` to confirm `nonBlank` semantics (trim → falsy collapses to `undefined`).
- Cross-referenced every consumer of the three affected paths:
  - `src/agents/cursor/index.ts:28,59,115,120` (`readIfExists`/`atomicWrite` on `AgentPaths.cursor.settingsJson`).
  - `src/agents/vscode/index.ts:23,26,37` (`readIfExists`/`atomicWrite` on `AgentPaths.vscode.mcpJson`).
  - `src/migrate/migrate.ts:147,184,292,330` (`AgentPaths.cursor.settingsJson` and `AgentPaths.vscode.mcpJson`).
  - None of `copilot.vscodeMcpInSettings`'s consumers were touched by the diff; the consumer surface is unchanged, the resolved value just becomes absolute when `APPDATA` is blank.
- Searched `src/` for any remaining `process.env.APPDATA` references. Only two sites remain: the new constant at `paths.ts:13` and the existing `resolveAgentSyncHome` site at `paths.ts:156`. Both are wrapped by `nonBlank`. No bare `?? ""` pattern is left.
- Verified `0598a11` (the referenced #117 commit) is in the repo history and addressed the same failure mode for `resolveAgentSyncHome`.
- Ran `bun test src/config/__tests__/paths.test.ts` → 26 pass, 0 fail.
- Ran a broader `bun test src/config/__tests__/paths.test.ts src/agents/cursor src/agents/vscode src/migrate` → 170 pass, 0 fail.
- Ran `bun run typecheck` (after `bun install`) → clean (`tsc --noEmit` exits 0).
- Ran `bun run lint` → 14 warnings + 1 info, all pre-existing in `scripts/*.ts` and `src/core/path-portability.ts`. No new findings in the two changed files.
- Scanned the changed files for the project's banned spec-tracker tokens (`FR-###`, `SC-###`, `US#`, `T###`, `NC-#`, `research R#`) per `CLAUDE.md`. None present.
- Inspected the CI matrix (`.github/workflows/ci.yml`, `e2e.yml`) to gauge how much regression protection the new tests actually buy.

## Findings

No findings. See Reasoning for the items I considered and rejected.

## Reasoning

A few things looked worth a second pass; I'm calling them out here so the rationale isn't invisible.

1. **`resolveAgentSyncHome` still uses an inline `nonBlank(process.env.APPDATA) ?? HOME` pattern at `paths.ts:156` rather than the new shared `APPDATA` constant.** I considered flagging this as a minor consistency nit, but the two are intentionally different: `resolveAgentSyncHome`'s fallback is `<HOME>` (yielding `<HOME>\agentsync`), whereas the new `APPDATA` constant falls back to `<HOME>\AppData\Roaming`. Folding them together would silently change the on-disk location of the AgentSync state dir on Windows hosts with a stripped `APPDATA` (from `<HOME>\agentsync` to `<HOME>\AppData\Roaming\agentsync`). That's a behavior change, out of this PR's stated scope, and arguably a separate question for a follow-up. Also, `resolveAgentSyncHome` reads at call time, so a module-level constant wouldn't fit without changing its testability contract. Correctly left alone.

2. **The new tests early-return on non-Windows, and this repo's CI matrix is Linux-only across `ci.yml` and `e2e.yml`.** I considered flagging the resulting CI-coverage gap. I'm not posting it because (a) the PR description explicitly discloses it ("the 3 new Windows-guarded tests short-circuit on Linux but still run their beforeEach/afterEach envelope"), (b) the alternative would require either adding a Windows runner to the CI matrix (out of scope, infra cost) or refactoring the win32 branch out of an IIFE into a pure helper testable on any host (also a separate change), and (c) the tests still serve as executable documentation and will catch regressions for anyone running the suite on Windows locally. Calling this a finding would be theater given the explicit disclosure.

3. **Cache-busting import in the new tests (`await import('../paths?v=${Date.now()}-${Math.random()}')`).** This is Bun-specific behavior; I confirmed it works because the test pass count went from 23 to 26 cleanly. Random-collision risk between concurrent reruns is negligible (`Date.now()` plus `Math.random()` to ~16 decimal places).

4. **The new tests assert `isAbsolute()` only, not the prefix `<HOME>\AppData\Roaming`.** A prefix check would be slightly stronger, but it's harder to write portably (path separators) and `isAbsolute` is the exact property that fails under the bug being fixed. Tight, targeted assertion. Fine as-is.

5. **`homedir()` itself returning `""` on a similarly stripped Windows environment.** If `USERPROFILE` is also blank, the new fallback `join("", "AppData", "Roaming")` is again relative. This is a theoretical hole, but (a) it's a strictly larger collapse than what this PR is addressing, (b) the pre-existing `resolveAgentSyncHome` has the same property, and (c) treating `homedir()` as authoritative is the convention everywhere else in the file. Not a regression introduced by this PR.

6. **No new bare `?? ""` introduced elsewhere.** Spot-checked: the diff only touches the three call sites named in the description, and the new constant uses `nonBlank` consistently.

Net: the fix is narrowly correct, the comment block clearly explains *why* the construct exists (so a future reader doesn't unwind it), and the tests pin the invariant on the platform where it matters. Recommend merge.

_cost: $1.2975 · turns: 26 · duration: 235s_

chrisleekr:

## Outstanding

Nothing blocks merge from the bot side. `reviewDecision` is not yet `APPROVED` on GitHub, so the human merge action is still pending; this is expected behavior (FR-017 — merge is a human action). No CI failures survived this iteration.

_cost: $0.4704 · turns: 8 · duration: 53s_

chrisleekr:
Superseded by #147 (merged 2026-05-29), which fixed the same issue #126 via the parameterized resolveWindowsAppData() helper. All three win32 APPDATA sites (cursor.settingsJson, copilot.vscodeMcpInSettings, vscode.mcpJson) now route through nonBlank on main. Closing as obsolete.

PR #175, opened 2026-06-10
ci: bust poisoned bun cache causing spurious node:fs/promises failures

chrisleekr:
## Root cause

The `Unit Tests` CI job started failing on Linux with `SyntaxError: Missing 'default' export in module 'node:fs/promises'` repeated across many test files at module load. After the previous PR (#173) merged, `main`'s Unit Tests check went red and stayed red.

**This is not a code, dependency, or Bun-version problem.** Evidence:

- The full unit suite passes **677 / 680** (3 unrelated env-specific) with **zero** `fs/promises` errors in a clean `linux/amd64 oven/bun:1.3.12` container built from a `git archive` of `HEAD`.
- Bun **1.3.14** (the version I first suspected) actually passes — the e2e `daemon-ipc` job is green on it. The failure also reproduces on the pinned **1.3.12**, so the Bun version is not the cause.
- No dependency does a default `import x from 'node:fs/promises'` (repo-wide grep clean); `@opentui/core` uses only named imports. The named module in the error is misleading.
- A passing run and the failing runs share the identical tree and lockfile — only the restored `actions/cache` entry differs.

The signature — a built-in module's export "vanishing" at load, only on the GitHub runner — is a **corrupted `~/.bun/install/cache` entry** restored by `actions/cache`. Bun then resolves modules from bad tarballs and emits the misleading error.

## Fix

- Add a `-v2` segment to the bun cache key (and its `restore-keys` prefix) in all four jobs, abandoning the poisoned cache namespace. First run repopulates clean.
- Correct the Bun-pin comments that wrongly blamed a 1.3.14 `fs/promises` regression.

If a poisoned entry ever recurs, the next step is dropping the dependency cache entirely (the install is ~10s for 52 packages).

## Verification

CI going green on this PR is the test of the hypothesis.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,18 @@ jobs:
       - name: Upgrade npm
         run: npm install --global npm@11.5.1
 
+      - name: Diagnose runtime (temporary — node:fs/promises investigation)
+        run: |
+          echo "== uname ==" && uname -a
+          echo "== os-release ==" && head -3 /etc/os-release
+          echo "== bun revision ==" && bun --revision
+          echo "== node ==" && node --version
+          echo "== minimal default import of node:fs/promises =="
+          printf 'import fs from "node:fs/promises"; console.log("readFile typeof:", typeof fs.readFile);\n' > /tmp/probe.mjs
+          bun /tmp/probe.mjs && echo "PROBE: default import OK" || echo "PROBE: default import FAILED"
+          echo "== single test file (focused) =="
+          bun test src/commands/__tests__/plugin.test.ts 2>&1 | head -20 || true
+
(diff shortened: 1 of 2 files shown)


chrisleekr:
Closing — investigation PR. Root cause CONFIRMED via CI diagnostics: not a Bun version bug, not actions/cache, not coverage, not our code. A minimal `import fs from 'node:fs/promises'` works on the runner, and the failing test file PASSES when run alone — but FAILS in the full `bun test` suite. This is Bun's shared module registry corrupting under many-file load on Linux, triggered/worsened by #153's @opentui/core 0.2.10→0.3.4 bump (heavy native+worker graph, value-imported by the TUI test files). Fix tracked separately.

